PR #11, opened 2025-02-14
Backend inclusion

Llannelongue:
I've tried to merge 3 different codebases (so high potential for carnage!): (1) the existing backend from `GA4HPC:v0.4_dashboard`, (2) the modifications to the backend by @ens-lgil to create the database, and (3) the Grafana frontend. 

Structure is fairly straightforward (and up for debate!):
- `backend/` : everything about the backend
- `database` : everything about the database linking backend and frontend
  - I've deleted the actual database file as it shouldn't be on github. The idea is for the database to be in `/database/db/`, which is part of the `.gitignore`.
- `frontend/` : everything about the frontend
- `data/` contains supporting data, such as servers infrastructure.  

I haven't had a chance to actually run any of the code yet, but I wanted to open this PR to give a chance to everyone to have a look. Feel free to add comments or merge in your own edits even! Maybe best way is for you to create a branch off this branch and submit your own PR, but happy to follow your lead on this.

[this is very much a draft PR but the option to add the "draft" tag is only for public repo]

- @ens-lgil: I've tried my best to include your modifications to the backend within the new folder structure, but I may have missed things! If you could have a quick look to see if it looks vaguely sensible to you
- `requirements.txt` is still outdated and needs to be recreated to account for both backend and frontend requirements
- I'll try to run the backend asap to check that it still works (using mock files). 
- It's ok if the frontend as is doesn't work perfectly, as this will be a WIP for @mattagape moving forward anyway.

It's a bit messy as a process, but hopefully we just have to do this once and then we can do things a bit more academically :)


frontend/import_dashboards.py
```diff
@@ -1,7 +1,7 @@
 import os,argparse
 import logging
-from grafana_ga.folder import GrafanaGAFolder
-from grafana_ga.dashboard import GrafanaGADashboard
+from frontend.grafana_ga.folder import GrafanaGAFolder
+from frontend.grafana_ga import GrafanaGADashboard
```
ens-lgil, line 4: Looks like it should be `from frontend.grafana_ga.dashboard ...` and I could see this in several scripts

ens-lgil (commented):
Seems to be fine for the merge as a starting point for the repo

mattagape:
Might be best to merge `main` into this branch first, and fix any issues arising. And *then* merge into `main`. This will protect the `main` branch from getting broken. Although, as it says there are no conflicts, this step is probably unnecessary.

frontend/import_dashboards.py
mattagape: In `import_dashboards.py`, the following works:
```
from grafana_ga.datasource import GrafanaGADataSource 
from grafana_ga.folder import GrafanaGAFolder
from grafana_ga.dashboard import GrafanaGADashboard
from grafana_ga.user import GrafanaGAUser
```

Llannelongue:
I’ve fixed the main issues in the backend arising from the refactoring, it can now run. I've also replaced the requirements files. The backend can for example be tested by running `backend/__init__.py` .

What I haven’t done:
- Anything to do with the database
- Anything to do with the frontend
- Add to `requirements.txt` the packages for the frontend (if any)

Agreed probably fine to merge as a starting point, and maybe save a copy of the legacy `main` branch (with working prototype for the frontend) to use as a reference if needed.

mattagape:
Normally, I think it's good to keep `main` as a correctly functioning branch, and branches should be working fully when merged into `main`, so that `main` doesn't get broken. But, in this case, I will merge the PR into `main`, and then work on fixing any changes.

In the meantime, before merging the PR, I have created a copy (to keep) of the `main` branch before merging this PR, which is called `original_main`.

PR #17, opened 2025-02-27
Feature/front end update

ens-lgil:
A first update which will certainly require more documentation and more error catching

### DIFF
--- a/frontend/grafana_ga/base.py
+++ b/frontend/grafana_ga/base.py
@@ -1,8 +1,13 @@
-
+import logging
+from niquests.exceptions import ConnectionError
 from grafana_client import GrafanaApi
 
 
+logger = logging.getLogger(__name__)
+
+
 class GrafanaGABase:
+    ''' Base class holding thea grafana_client GrafanaApi instance, the Grafana URL and the admin login/password'''
 
     def __init__(self, login:str, password:str, grafana_url:str) -> None:
         base_url = "http://{}:{}@{}".format(login, password, grafana_url)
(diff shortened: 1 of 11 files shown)


frontend/grafana_ga/base.py
```diff
@@ -1,8 +1,13 @@
...
+import logging
+from niquests.exceptions import ConnectionError
 from grafana_client import GrafanaApi
 
 
+logger = logging.getLogger(__name__)
+
+
 class GrafanaGABase:
+    ''' Base class holding thea grafana_client GrafanaApi instance, the Grafana URL and the admin login/password'''
```
mattagape: thea -> the ?

requirements_frontend.txt
```diff
@@ -0,0 +1,2 @@
+psycopg[binary]==3.2.5
```
mattagape: Do we definitely want the precise versions, or should we put these as minimum versions? I'm not sure what people normally do with this sort of thing.

frontend/grafana_ga/folder.py
```diff
@@ -7,6 +7,7 @@
 
 
 class GrafanaGAFolder(GrafanaGABase):
+    ''' Class used to create Grafana dashboard folder an setup the teams/users permissions on it '''
```
mattagape: an -> and
teams/users -> teams'/users'

mattagape (changes requested):
A couple of trivial changes, plus a question on versioning.

requirements_frontend.txt
ens-lgil: Yes, we could to that instead. But this is mainly to provide a version that we know works correctly with the code

frontend/grafana_ga/folder.py
ens-lgil: Thanks! I missed these typos

requirements_frontend.txt
mattagape: Fair enough

mattagape (approved):
Thanks for those changes. I think this PR will be really helpful.

mattagape:
Go ahead and merge

PR #19, opened 2025-02-27
Matt backend

mattagape:
This is the beginning of the new `ga_dashboard` module, and basically is the backend moved into the module structure, and the addition of `pytest`-compatible tests (very basic at the moment).

For the module to be used properly, it will need more information, either a `setup.py` or `.toml` file, I think.

Later, we will move the frontend into the module as well.


frontend/grafana_ga/folder.py
```diff
@@ -1,4 +1,30 @@
...
+
+#  in code: 
+# ['/Users/mg2216/repos/GA4HPCdashboard/frontend/grafana_ga', 
+# '/opt/miniconda3/envs/py313/lib/python313.zip', both
+# '/opt/miniconda3/envs/py313/lib/python3.13',   both
+# '/opt/miniconda3/envs/py313/lib/python3.13/lib-dynload', both
+# '/opt/miniconda3/envs/py313/lib/python3.13/site-packages']  both
+
+
+import sys
```
ens-lgil: Maybe this block can be removed or commented as it looks like this is only for debugging

ga_dashboard/LICENSE
ens-lgil, line 1: Wondering if we should place the LICENSE, CONTRIBUTING, CHANGELOG and README at the root of the repository.
But I guess it depends how we will set the repository at the end (i.e. if the ga_dashboard will be the root of the repository)

ga_dashboard/run_backend.sh
```diff
@@ -14,7 +17,7 @@ userCWD="$(pwd)"
 parent_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 cd "$parent_path"
 
-# The only requirement for the module is to start python 3.11 or later
+# Module loading line. The only requirement for the module is to start python 3.11 or later
```
ens-lgil, line 20: Should we set the python minimal version to 3.11 or more ?
In GreenAlgorithms4HPC the python minimal version is relatively old (v3.8): [README](https://github.com/GreenAlgorithms/GreenAlgorithms4HPC/blob/main/README.md)

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
ens-lgil, line 1: It's good for now, but in the future it might make more sense to use the `logging` package, like in the **grafana_ga** modules.

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
```diff
@@ -185,6 +198,10 @@ def enrich_data(df:pd.DataFrame, fParams:dict, users_df:pd.DataFrame, GA:GA_tool
 
 def summarise_data(df:pd.DataFrame, args:argparse.Namespace) -> dict:
 
+    if ( df is None ):
```
ens-lgil: Minor thing: I think the parenthesis are not necessary in the `if` statements.

ga_dashboard/src/ga_dashboard/backend/utils.py
```diff
@@ -71,7 +77,8 @@ def _validate_db_conn(self, args):
             )
             conn.close()
         except psycopg.OperationalError as err:
-            raise(f'/!\ Error: Issue to connect to the database: {err}')
+            raise(f'Error: Issue to connect to the database: {err}')
+            #raise(f'/!\ Error: Issue to connect to the database: {err}')
```
ens-lgil: Commented line can be removed

ens-lgil (commented):
Looks good to me at this stage!

frontend/grafana_ga/folder.py
mattagape: Removed!

ga_dashboard/LICENSE
mattagape: Hi. I think I'll leave them in place for now. I have tried to follow this tutorial:
https://py-pkgs.org
but without using tools like `poetry`, which I've not used before.
That tutorial develops a simple package, which you can see here:
https://github.com/TomasBeuzen/pycounts
I envisage that the front-end and backend code will all finally live in a package called `ga_dashboard`. So I think these  meta-data files are OK to stay where they are for now.

ga_dashboard/run_backend.sh
mattagape: Do you think we should change it to v3.8, then? It's currently 3.11 in `run_backend.sh`.

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
mattagape: Indeed. I've added a `TODO` in this file to remind me to do so.

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
mattagape: Removed!

ga_dashboard/src/ga_dashboard/backend/utils.py
mattagape: Removed!

ga_dashboard/run_backend.sh
ens-lgil: We need to see with @Llannelongue , but 3.11 should be better as the IT systems use supported versions of Python and the 3.8 is not supported anymore: https://devguide.python.org/versions/

mattagape: In that case, we should definitely use a later version than 3.8!

ens-lgil (approved):

ga_dashboard/run_backend.sh
Llannelongue: I guess there are two aspects here: (1) what is it built for, and (2) what is it compatible with. 
For (1), it makes sense to use v3.11 to have the latest version of Python, and similarly use the latest version of all the packages used. It also avoids bugs of older versions. 

For (2), we want to maximise compatibility in case an HPC platform only has for example 3.10. 

I think it makes sense to work with 3.11 so far. But before releasing, we can try with older Python versions and check which ones work as well?

mattagape: Yes. Judging by the list of supported versions, I think the oldest we should support would be 3.9, which appears to be the oldest version still supporting security updates. But sure, before release we can check this

ga_dashboard/src/ga_dashboard/backend/extract/slurm.py
```diff
@@ -330,14 +330,14 @@ def clean_logs_df(self):
         if 'CPUTime' in self.logs_df.columns:
             self.logs_df['CPUwallclocktime_'] = self.logs_df['CPUTime'].apply(self.parse_timedelta)
         else:
```
Llannelongue: This second condition can now be removed, it was just to account for older versions of GA4HPC.
The only line to keep is:
`self.logs_df['CPUwallclocktime_'] = self.logs_df['CPUTime'].apply(self.parse_timedelta)`

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
```diff
@@ -45,20 +46,26 @@
...
-        Calculate the energy usaged based on the job's paramaters
+        Calculate the energy usage based on the job's parameters
         :param row: [pd.Series] one row of usage statistics, corresponding to one job
         :return: [pd.Series] the same statistics with the energies added
         '''
         ### CPU and GPU
-        partition_info = self.cluster_info['partitions'][row.PartitionX]
+        try:
+            partition_info = self.cluster_info['partitions'][row.PartitionX]
+        except KeyError as ke:
```
Llannelongue, line 62: What's the motivation behind this exception?

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
```diff
@@ -147,29 +155,34 @@ def extract_data(args:argparse.Namespace, has_slurmAdmin:bool, cluster_info) ->
...
+
+def enrich_data(df:pd.DataFrame, fixed_params:dict, users_df:pd.DataFrame, GA:GA_tools) -> pd.DataFrame:
 
     ### energy
     df = df.apply(GA.calculate_energies, axis=1)
 
-    df['energy_failedJobs'] = np.where(df.StateX == 0, df.energy, 0)
+    try:
+        df['energy_failedJobs'] = np.where(df.StateX == 0, df.energy, 0)
+    except AttributeError as err:
```
Llannelongue, line 169: Same question here, would be good to add a comment to explain what can cause this issue (and how to address it)

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
```diff
@@ -348,18 +367,16 @@ def import_data_in_db(summary_stats:dict,args) -> None:
...
+                           'startDay endDay useCustomLogs use_mock_agg_data reportBug reportBugHere path_infrastructure_info')
     args = argStruct(
         startDay='2022-01-01',
         endDay='2023-06-30',
         useCustomLogs="", #"sacct_output_loic1.txt",
         use_mock_agg_data=True,
         reportBug=False,
         reportBugHere=False,
-        path_infrastucture_info="data/ourInfrastructure/CSD3",
+        path_infrastructure_info="../../../data/ourInfrastructure/CSD3", # change?
```
Llannelongue: Probably shouldn't have the name/path of our test infrastructure in the file (since these aren't on the GitHub on purpose)

Llannelongue:
Re `ga_dashboard/src/ga_dashboard/backend/example_files`: does it make sense to have these example files there still? They were meant as a debugging example for users, and probably worth keeping, but maybe not nested in the package like this? (as they're not meant to be run or used by the code). They could almost be part of the doc

Llannelongue (approved):
Looks fine thanks, I've just added a few comments that can be addressed before merging

mattagape:
> Re `ga_dashboard/src/ga_dashboard/backend/example_files`: does it make sense to have these example files there still? They were meant as a debugging example for users, and probably worth keeping, but maybe not nested in the package like this? (as they're not meant to be run or used by the code). They could almost be part of the doc

Yes. The aim at the moment is to get the existing code, frontend and backend, into a package structure. Once that is done, we can start adding tests, keeping/removing files, adding to the documentation, etc. One step at a time!

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
mattagape: Quite right. Removed!

ga_dashboard/src/ga_dashboard/backend/extract/slurm.py
mattagape: Updated.

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
mattagape: The motivation was that I was just trying to get this script to work. Obviously we will want to make our code as robust as possible, and raise exceptions when information is missing/incomplete/corrupted

mattagape: Comment and `TODO` added to code.

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
mattagape: Comment and `TODO` added to code.

mattagape:
@Llannelongue I've made a couple of requested changes. Other things will need more work. At the moment, the backend code on this branch is still being worked on. But we have the working version safely stored in  the `original_main` branch.
This is not ideal. Ideally, changes which break `main` should never be merged in. If you want, I can spend more time on the `matt_backend` branch before completing the PR. Or, we can merge now, and I can address this later.
I think the issue is the command arguments etc

mattagape:
OK, I've had a quick look. I think the issue is missing test data files on my machine (or in wrong place). So, I think we can go ahead and merge.

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
ens-lgil: This could an additional thing to build: a check that the cluster_info file contains all the required field (check format ?)
And we could also do the same for the Slurm data file(s).

mattagape: Yes, @ens-lgil. That sort of thing is definitely what we will want to check for, and make sure the code is working properly (and catching missing fields) with the unit tests.

Llannelongue: That'd work. Also if this bit of code is just to make the code run for now but not a long-term test, it'd be could to highlight that with a comment so that we can keep track. I've been using the `# DEBUGONLY` tag for that to quickly identify line that can be removed, but any system would work.

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
Llannelongue: (same comment as above regarding debug only code)

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
mattagape: Well, eventually we'll ideally check for missing keys every time we try to access them, so I'll leave it as-is for now.

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
mattagape: as above.

ga_dashboard/src/ga_dashboard/backend/ga_tools.py
Llannelongue: Doesn't it introduce a lot of unnecessary checks? If we know we've just created the fields at a previous step, we don't need to check that it's there? Only when there is the option of a missing field it's worth checking.

mattagape: They are not unnecessary. I added them because the code kept crashing at that point.

In any case, it's best to make our code as robust as we can, and fail early with meaningful information to the user.

PR #24, opened 2025-03-19
Pathfix

mattagape:
Mainly fixing some things and adding more documentation.


requirements_frontend.txt
```diff
@@ -1,2 +1,3 @@
 psycopg[binary]>=3.2.5
+psycopg2>=2.9.10
```
ens-lgil: I would recommend to use `psycopg`**3** instead (i.e. psycopg[binary]>=3.2.5).
It is used in frontend/import_mockup_aggregate.py (by replacing `import psycopg2` by `import psycopg`).
Django recommmends to use the version 3 (cf [doc](https://docs.djangoproject.com/en/5.1/ref/databases/#postgresql-notes)) and is likely to deprecate de version 2 so it might be good to follow their lead.

mattagape: Thanks. The requirement for `psycopg2` was in `import_mockup_aggregate.py`. I've uninstalled `psycopg2` and changed this to `psycopg` in the module import and the place it was used in that script, and it seems ok

ens-lgil (approved):

Llannelongue (approved):
Looks fine

PR #32, opened 2025-04-01
End to end

mattagape:
Our first end-to-end demo!


demo.sh
```diff
@@ -0,0 +1,101 @@
...
+# --reportBug for run_backend.sh
+# --reportBugHere for run_backend.sh
+# --useCustomLogs for run_backend.sh
+datasource_name=$DEFAULT_DATASOURCE_NAME
+grafana_url=$DEFAULT_GRAFANA_URL
+common_users_file=$DEFAULT_USERS_FILE
+sacct_file=$DEFAULT_SACCT_FILE
+
+echo "\n*** Importing users to Grafana: ***\n" # This step will fail if Grafana is not running.
+python $repo_root_dir/frontend/import_users.py --input_file $common_users_file \
```
ens-lgil, line 82: The parameter `--url $grafana_url` seems missing here

frontend/grafana_ga/user.py
```diff
@@ -27,33 +27,42 @@ def create_team(self, team_name:str) -> None:
...
 
-        # # Add team ID <-- looks like redundant code with the "if ex.status_code == 409:" code block
-        # try:
-        #     teams_list = self.grafana.teams.get_team_by_name(team_name)
-        #     if teams_list and len(teams_list) > 0:
-        #         uid = teams_list[0]['uid']
-        #         self.teams[uid] = teams_list[0]
-        # except GrafanaClientError as ex:
-        #     logger.error(f"ERROR during fetching of team '{team_name}': {ex}")
+        # Add team ID
```
ens-lgil, line 45: Is it not redundant anymore ?

demo.sh
```diff
@@ -0,0 +1,101 @@
...
+infrastructure_dir=$DEFAULT_INFRASTRUCTURE_DIR
+# --reportBug for run_backend.sh
+# --reportBugHere for run_backend.sh
+# --useCustomLogs for run_backend.sh
+datasource_name=$DEFAULT_DATASOURCE_NAME
+grafana_url=$DEFAULT_GRAFANA_URL
+common_users_file=$DEFAULT_USERS_FILE
+sacct_file=$DEFAULT_SACCT_FILE
+
+echo "\n*** Importing users to Grafana: ***\n" # This step will fail if Grafana is not running.
```
ens-lgil, line 81: Minor stuff but this `echo` gives a strange output format in my shell:
```
\n*** Importing users to Grafana: ***\n
```
And I got the same type of output format for each `echo` in the script.

frontend/import_mockup_aggregate.py
```diff
@@ -32,18 +39,41 @@ def get_columns(input_file):
...
     db_col_names = {}
     for file_col in file_col_names:
+
+    # Out of User, UID, Name, Group and Department we now only store User in the ga_data_aggregate table
+
+        if user_data:
+            if file_col not in ['User', 'UID', 'Name', 'Group', 'Department']:
+                continue
+
+        if not user_data:
```
ens-lgil: Maybe we could replace it simply by an `else:`

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
```diff
@@ -100,15 +101,42 @@ def import_data(self) -> None:
...
+                # data is a list of dictionaries; item is one such dictionary
+                # NB there are multiple, identical instances of each data item. We only want the first in each. 
+                for item in data:
+
+                    # As we no longer want all columns in the data, we need to specify each one we want.
+                    # See https://discuss.python.org/t/how-to-do-multi-line-f-strings/46634/6 - avoiding SQL injection attacks.
+                    # and https://www.psycopg.org/psycopg3/docs/basic/params.html
+                 
+                    # The ON CONFLICT DO NOTHING is a temporary hack to stop the duplicate values being (attempted to be) added 
+                    sql = "INSERT INTO ga_data_aggregate (user_name, submitdate, n_jobs, first_job_period, last_job_period, energy," \
```
ens-lgil: Not sure this would be the cleanest way to build the SQL query.
We could use the list of column names to build:
- the list of columns
- the list of values

frontend/grafana_ga/user.py
mattagape: That's right. I should not have removed it!

demo.sh
mattagape: Is `set -x` not commented out (line 7)

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
mattagape: The original version just used all the columns in the data file. We don't need all of them, so I had to specify which ones I wanted.

Do you have another idea in mind?

demo.sh
mattagape: Oops! Thanks for pointing that out. I'll add it.

demo.sh
ens-lgil: `set -x` is commented in demo.sh

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
ens-lgil: Just creating a list of the columns we want and then generate a list from the `item` dictionnary so in the future we would have to change the list of columns in one place, e.g. something like:
```
columns = ['user_name', 'submitdate', 'n_jobs', 'first_job_period', 'last_job_period', 'energy', 'energy_cpus', 'energy_gpus', 'energy_memory', 'carbonfootprint', 'carbonfootprint_memoryneededonly', 'carbonfootprint_failedjobs', 'cputime', 'gputime', 'wallclocktime', 'cpuhourscharged', 'gpuhourscharged', 'memoryrequested', 'memoryoverallocationfactor', 'n_success', 'treemonths', 'treemonths_memoryneededonly', 'treemonths_failedjobs', 'driving', 'flying_ny_sf', 'flying_par_lon', 'flying_nyc_mel', 'cost', 'cost_failedjobs', 'cost_memoryneededonly', 'success_rate', 'failure_rate', 'share_carbonfootprint']

data = []
for col in columns:
	if col in item.keys()
		val = item[col]
		# Check if numeric 
		if re.match(r'^\d+\.?\d+$,str(val)):
			data.append(val)
		else:
			data.append(f"'{val}'")


sql = f"INSERT INTO ga_data_aggregate ('{"','".join(columns)}') VALUES({','.join(data)}) ON CONFLICT DO NOTHING"
```

demo.sh
```diff
@@ -65,6 +70,13 @@ grafana_url=$DEFAULT_GRAFANA_URL
 common_users_file=$DEFAULT_USERS_FILE
 sacct_file=$DEFAULT_SACCT_FILE
 
+echo "\n*** Setting up Postgres database: ***\n"
+export PGPASSWORD="$db_password"
+psql -c 'drop database if exists ga_db; ' -U postgres
+psql -c 'create database ga_db; ' -U postgres
+psql -U $db_user -d ga_db < $repo_root_dir/database/ga_db.sql ; export PGPASSWORD=
```
ens-lgil: We might need to add the port, just in case

demo.sh
mattagape: Which shell are you using, and how do you invoke the script?

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
mattagape: Yes, your way is much better :) I'll change it.

demo.sh
mattagape: Good point. I've just added the port and host to this part.

frontend/import_mockup_aggregate.py
mattagape: Indeed we could! Changed.

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
mattagape: It doesn't like your re:
```
Traceback (most recent call last):
  File "/Users/mg2216/repos/GA4HPCdashboard/ga_dashboard/run_backend.py", line 8, in <module>
    from ga_dashboard.backend.ga_tools import main_backend
  File "/Users/mg2216/repos/GA4HPCdashboard/ga_dashboard/src/ga_dashboard/backend/ga_tools.py", line 12, in <module>
    from ga_dashboard.backend.data_sql_import import DataSQLImport
  File "/Users/mg2216/repos/GA4HPCdashboard/ga_dashboard/src/ga_dashboard/backend/data_sql_import.py", line 144
    if re.match(r'^\d+\.?\d+$,str(val)):
                ^
SyntaxError: unterminated string literal (detected at line 144)
```

mattagape: OK I think I've fixed it

mattagape: Your way is simpler, but I think could be open to SQL injection attacks. See https://www.psycopg.org/psycopg3/docs/basic/params.html#danger-sql-injection

ens-lgil: There should be an hybrid solution where we could set the number of `%s` in the VALUES and use the list of data as `params` in a safe way

demo.sh
ens-lgil: I used the default `zsh` shell in Mac, but I used the `bash` command to launch the script

mattagape: try `sh ./demo.sh password` or `zsh ./demo.sh password`

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
mattagape: Yes, I'm just looking at it

mattagape: Done.

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
```diff
@@ -111,25 +112,41 @@ def import_data(self) -> None:
...
-                        item['carbonfootprint_failedjobs'], item['cputime'], item['gputime'], item['wallclocktime'], item['cpuhourscharged'], \
-                        item['gpuhourscharged'], item['memoryrequested'], item['memoryoverallocationfactor'], item['n_success'], \
-                        item['treemonths'], item['treemonths_memoryneededonly'], item['treemonths_failedjobs'], item['driving'], \
-                        item['flying_ny_sf'], item['flying_par_lon'], item['flying_nyc_mel'], item['cost'], item['cost_failedjobs'], \
-                        item['cost_memoryneededonly'], item['success_rate'], item['failure_rate'], item['share_carbonfootprint'])
-                    
-                    cur.execute(sql, params)
+                    #sql = f"INSERT INTO ga_data_aggregate ('{"','".join(columns)}') VALUES({','.join(data)}) ON CONFLICT DO NOTHING"
+                    sql = f"INSERT INTO ga_data_aggregate ({", ".join(columns)}) VALUES("
+                    sql += "%s"
```
ens-lgil, line 143: A slightly simpler way:
```
for i in range(len(columns)):
    sql += "%s"
```
And then we could remove the line `num_columns = len(columns)` and the first `sql += "%s"`

mattagape: I think we need a comma between each two instances of `%s`

ens-lgil: Should we update the code before the merge ?

frontend/grafana_ga/user.py
ens-lgil: Should we comment/remove the code before the merge ?

mattagape: No, we need to leave it in. Otherwise the script breaks when you first try to add teams etc. to Grafana.

ga_dashboard/src/ga_dashboard/backend/data_sql_import.py
mattagape: I already have?

mattagape: Unless you mean the redundant comments? I could remove those

mattagape: To handle the commas, we need to use `range(1, num_columns)`, i.e., start with index 1.

ens-lgil: Yes.
I just had in mind something like this to keep everything within the loop:
```
if i == 0:
  sql += "%s"
else:
  sql += ", %s"
```
or even more compact:
```
sql += "%s" if i==0 else ", %s"
```
But you code works so we can go with it.

Llannelongue:
It'd be good to clarify the steps to installing postgresql in a way that works for this, as it's not immediately obvious from their documentation

mattagape:
> It'd be good to clarify the steps to installing postgresql in a way that works for this, as it's not immediately obvious from their documentation

I've added a note about this to the README.

Issue #31, opened 2025-03-26
ga_dashboard package folder structure.


mattagape:
The current folder structure for the `ga_dashboard` package is confusing:  e.g., `GA4HPCdashboard/ga_dashboard/src/ga_dashboard/frontend/`.

It would be good if this could be simplified.

mattagape:
Now simplified.

mattagape:
Re-opening until PR merged.

PR #36, opened 2025-04-28
Packaging

mattagape:
The scripts run OK in an asynchronous manner. I have updated the documentation and have written part of a Worked Example to show how Eq (1) is used in the 2021 Green Algorithms paper.


scripts/backend/run_sacct_only.py
```diff
@@ -0,0 +1,95 @@
...
+        args.startDay,  # format YYYY-MM-DD
+        "--endtime",
+        args.endDay,  # format YYYY-MM-DD
+        "--format",
+        "UID,User,JobID,JobName,Submit,Elapsed,Partition,NNodes,NCPUS,TotalCPU,CPUTime,ReqMem,MaxRSS,WorkDir,State,Account,AllocTres",
+        "-P",
+        "-L"  # All clusters
+    ]
+
+    filename = args.outFile
```
ens-lgil, line 66: Maybe we could add a check to make sure that the output directory exists

scripts/end-to-end/demo.sh
```diff
@@ -69,37 +70,39 @@ datasource_name=$DEFAULT_DATASOURCE_NAME
...
 psql -c 'drop database if exists ga_db; ' -U postgres -h $db_host -p $db_port
 psql -c 'create database ga_db; ' -U postgres -h $db_host -p $db_port
-psql -U $db_user -h $db_host -p $db_port -d ga_db < $repo_root_dir/database/ga_db.sql
+psql -U $db_user -h $db_host -p $db_port -d ga_db < $repo_root_dir/ga_dashboard/database/ga_db.sql
 export PGPASSWORD=
 echo "\n* Done! *\n"
 
 echo "\n*** Importing users to Grafana: ***\n" # This step will fail if Grafana is not running.
-python $repo_root_dir/frontend/import_users.py --input_file $common_users_file \
+python $repo_root_dir/scripts/frontend/import_users.py --input_file $common_users_file \
```
ens-lgil: Could be useful to setup variables for the `frontend` and `backend` script directories as they are used more than once in the file, e.g.:
```
frontend_script="$repo_root_dir/scripts/frontend/"
backend_script="$repo_root_dir/scripts/backend/"
```

mattagape: Indeed. I've changed it in cefd2b6.

scripts/backend/run_sacct_only.py
mattagape: The output is just collected in the directory where the command is invoked, so the directory must exist?

ens-lgil (approved):

ga_dashboard/backend/extract/slurm.py
```diff
@@ -1,11 +1,11 @@
-import subprocess
 import argparse
```
Llannelongue: (minor point) Would it make sense to rename `/backend/extract` into something clearer, maybe `/backend/WorkloadManager_extract`?

ga_dashboard/backend/extract/slurm.py
```diff
@@ -235,7 +239,7 @@ def __init__(self, args:argparse.Namespace, cluster_info):
 
         try:
             self.args_dict = self.args.__dict__  # This is when using command line arguments (Namespace)
-        except:
+        except Exception:
```
Llannelongue: Out of curiosity, what's the motivation behind this change?

ga_dashboard/backend/ga_tools.py
```diff
@@ -165,7 +163,9 @@ def extract_data(args:argparse.Namespace, has_slurmAdmin:bool, cluster_info) ->
     return WM.df_agg_X
 
 
-def enrich_data(df:pd.DataFrame, fixed_params:dict, users_df:pd.DataFrame, GA:GA_tools) -> pd.DataFrame:
+def enrich_data(df: pd.DataFrame, fixed_params: dict, users_df: pd.DataFrame, GA: GA_tools) -> pd.DataFrame:
+
+    ## FIXME? Calling this will fail if not all users are in the users_df
```
Llannelongue: Shall we add a `try` to catch this error when it happens?

Llannelongue:
`ga_dashboard/database/import_aggregate_data.py` -> is this still part of the pipeline or is it a legacy for debugging?

ga_dashboard/samples/grafana_users_list.csv
```diff
@@ -1,4 +1,4 @@
-Name,User name,Email,Password,Group
+Name,User,Email,GrafanaPassword,Group
```
Llannelongue: Not an issue for the test data, but do we recommend IT teams to store the list of passwords in clear like this? Or what would be a best practice we could recommend?

scripts/RunningScripts.md
```diff
@@ -0,0 +1,93 @@
+# Running the scripts
```
Llannelongue, line 1: The `/script/` folder is a bit cryptic (everything is this repo are scripts), what could be better names for it?

scripts/RunningScripts.md
```diff
@@ -0,0 +1,93 @@
...
+
+The scripts in these directories can be used to run the entire cycle:
```
Llannelongue, line 7: Could we clarify in this document the difference between setting up the dashboard the first time (one-off), and the ongoing deployment. Then being clear about the difference between backend and frontend.

scripts/RunningScripts.md
```diff
@@ -0,0 +1,93 @@
...
+
+1. Initialise the backend database
+2. Run `sacct` command, enhance the data and load it into the databse
```
Llannelongue: Perhaps better:
2. Running the backend (run `sacct` command, enhance the data and load it into the database)

scripts/RunningScripts.md
```diff
@@ -0,0 +1,93 @@
...
+3. Set up the dashboards, users, etc.
+
+It should then be possible to view the data in the dashboards running on the Grafana server.
+
+## Backend database set-up
```
Llannelongue: Throughout the project, we seem to sometimes including the database in the backend, and sometimes it's a standalone component alongside backend and frontend. Which one should it be?

scripts/RunningScripts.md
```diff
@@ -0,0 +1,93 @@
...
+Use `init_db.sh`.
+
+Example:
+```
+sh scripts/backend/init_db.sh <password>
+```
+where `<password>` in this case corresponds to the back-end `postgres` user password.
+
+<ins>**WARNING!! This script will delete any existing instance of the database, including all its data, so only run this if you are sure that's what you want.**</ins>
```
Llannelongue, line 40: This seems very risky. Can we not have some kind of safety mechanism? Either asking to confirm deletion, or adding a flag to `init_db,sh` so that it crashes if there is an existing database and the flag isn't used. Or other option to protect the database.

scripts/RunningScripts.md
```diff
@@ -0,0 +1,93 @@
...
+
+As well as setting up the database and its tables, the script also adds HPC users to the database; alternatively, you can run the `add_users_to_database.py` script directly. Note that we want to add HPC users to the database before running `sacct`; the users we want should exists in the backend database regardless of whether there is any `sacct` data for them!
+
+
+## Extract data with `sacct`, and enrich it
+Assuming you **cannot** run the code for the database, etc., on the HPC machine, you will need to run just the `sacct` command
```
Llannelongue: As discussed, this is a very niche case, and shouldn't be the first thing we advertise for now.

scripts/RunningScripts.md
```diff
@@ -0,0 +1,93 @@
...
+* The options for the start and end dates, `-S` and `-E`, are not needed, as the date range will be determined by the contents of the `sacct` output file.
+* The directory given with the `--useOtherInfrastructureInfo` option must contain **both** the cluster information file (`cluster_info.yaml`) and the file listing the HPC users (`hpc_users_list.csv`); here, the `samples` subdirectory, which is part of the repo, is used.
+* In this example, we use the file `sacct_20250414.txt`, which you just generated. Sample `sacct`-output files are also in the `samples` subdirectory. 
+
+**Alternatively, if you are able to run everything on the HPC machine**, invoke the `run_backend.sh` script WITHOUT the `--useCustomLogs` option. For example:
+
+```
+sh scripts/backend/run_backend.sh \
+    --db_name ga_db --db_user postgres --db_password <password> 
+    -S 2025-04-14 -E 2025-04-15 \
```
Llannelongue: The one-day window is only for daily runs, not for the usual ones

scripts/RunningScripts.md
```diff
@@ -0,0 +1,93 @@
...
+    --useOtherInfrastructureInfo ~/repos/GA4HPCdashboard/ga_dashboard/samples \
+    --fixed_params_file ~/repos/GA4HPCdashboard/ga_dashboard/data/fixed_parameters.yaml
+```
+
+This will extract usage information using `sacct`, enhance it (work out some additional quantities using it), then add it to the database. Obviously, you will need to change the values passed as the script arguments to your own set-up. You will also need the `-S` and `-E` options, as shown above.
+
+## Grafana set-up
+A number of things need to be done, assuming you have downloaded Grafana and are able to run its server. These are:
+* import the dashboard(s) you want to use into the Grafana server
```
Llannelongue, line 102: As a default, we don't want to make it too complicated here: we give them a dashboard to use. Only if they want to divert from that do they have to do something different.

scripts/backend/add_data_to_db.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
```
Llannelongue: It's a lot of different files in `/scripts/`, are they still all needed, or are some legacy ones, or maybe belonging in `/ga_dashboard`? 

I guess on the same line, what is the distinction between what goes in `/ga_dashboard` and what goes in `/scripts/`?

mattagape:
> `ga_dashboard/database/import_aggregate_data.py` -> is this still part of the pipeline or is it a legacy for debugging?

I think it's legacy. It doesn't seem to be used in the pipeline. And, it uses a table called `ga_data_aggregate`, which isn't in our database. So I'll delete it.

Removed in 608f04a.

mattagape:
This PR fixes #31

Llannelongue (changes requested):

ga_dashboard/backend/extract/slurm.py
mattagape: It was flagged by the `ruff` linter: https://docs.astral.sh/ruff/rules/bare-except/

ga_dashboard/backend/extract/slurm.py
mattagape: I don't think we'd gain anything doing that. And it's probably best not to name the directory after the Python class, in case the latter changes.

ga_dashboard/samples/grafana_users_list.csv
mattagape: We absolutely wouldn't want to have them in the clear like that! I believe passwords aren't normally stored anyway, just their hash.

scripts/RunningScripts.md
mattagape: The `scripts` folder is meant to be programs which use the `ga_dashboard` package, but which aren't themselves part of that package. Thus, as well as being useful, they show examples of how to use the package. I suppose you could call them client code (clients of the package).

ga_dashboard/backend/extract/slurm.py
Llannelongue: The name could be something else, doesn't have to be the same as the class. The gain is clarity as "extract" could mean different things and e.g. we don't extract from the database in this. What this folder contains is everything that is **workload manager-specific**, so it could be good to name it accordingly.

scripts/RunningScripts.md
mattagape: It's used by both the backend and the frontend, so I suppose we should refer to it as standalone.

scripts/RunningScripts.md
mattagape: Indeed. Changed in 26126fa.

ga_dashboard/samples/grafana_users_list.csv
Llannelongue: Would it make sense then to have the test files similar to what we want the real thing to be? We can have the passwords in the README anyway for information, but they're not needed in the user_list

ga_dashboard/backend/extract/slurm.py
mattagape: Fair enough. We could call it `/workload_manager/`, perhaps, instead of `/extract/`?

Llannelongue: Sounds good

scripts/RunningScripts.md
Llannelongue: Ok then can we tweak the text and structure to reflect that?

ga_dashboard/backend/ga_tools.py
mattagape: Hmm, well we already raise a `ValueError` on line 200 if that occurs.

scripts/RunningScripts.md
mattagape: I thought it would operate as a daily `cron` job?

ga_dashboard/backend/extract/slurm.py
mattagape: Changed in 157ab93.

ga_dashboard/samples/grafana_users_list.csv
mattagape: The passwords are used if one adds users to Grafana in a script.
See `set_new_user()` function in `ga_dashboard.grafana_ga.user.py`

mattagape: I'm not sure how one would store the hashes instead. I'd need to look into it.

scripts/RunningScripts.md
mattagape: Addressed this one in d71a30f.

scripts/RunningScripts.md
mattagape: Yes. Addressed in 3c776d8.
Now the script has been renamed to make it more explicit. It also prompts the user for the password, rather than having it as a command-line option.

ga_dashboard/backend/ga_tools.py
mattagape: Although I'm not sure about this line:
```
if len(df2) != len(df):
```
What if you have users in the database but no `sacct` data for them? That shouldn't fail. I need to write some tests for this one, I think, to check what happens.

scripts/RunningScripts.md
mattagape: I've simplified the relevant script, so that it uses more default options.
See abd223e
NB Some scripts need absolute path names for their arguments for input files (etc.), whilst others are OK with relative paths. It would be good to make this consistent. Shall I put this as a new issue, or do you want it fixed now?

scripts/backend/add_data_to_db.sh
mattagape: I've removed one script. See 029643e.
I could remove the `run_sacct_only.py` script, but it does potentially save some typing. See c9a41f3

scripts/RunningScripts.md
mattagape: On second thoughts, I think it's OK as-is, as users will probably want to refer to paths outside the repo, and so need absolute paths.

scripts/backend/add_data_to_db.sh
mattagape: The `/scripts/` folder is meant to be programs which use the `ga_dashboard` package, but which aren't themselves part of that package. Thus, as well as being useful, they show examples of how to use the package. I suppose you could call them client code (clients of the package).

scripts/RunningScripts.md
mattagape: See also 7ad4134.

scripts/RunningScripts.md
mattagape: Probably best to do that on another branch?

ga_dashboard/backend/ga_tools.py
mattagape: I've now done a couple of little tests, and have added an explanatory comment where the error is raised. I think this one has therefore been dealt with. Please see cbf4472.

ga_dashboard/samples/grafana_users_list.csv
mattagape: I think this should be looked at separately to this PR. See new issue #37.

scripts/RunningScripts.md
mattagape: For, now put in a new issue. Please see #38

scripts/RunningScripts.md
mattagape: Please see commit 2fb55ad and issue #38, respectively.

scripts/RunningScripts.md
Llannelongue: Yes except the first install where it needs to run for a longer period. It would be good to clarify the two cases in the differences (if it hasn't been addressed already by another commit).

scripts/RunningScripts.md
Llannelongue: > It also prompts the user for the password, rather than having it as a command-line option.

What is the motivation behind this?

mattagape: To make it less likely that the user will run the script without thinking, and inadvertently delete all their data. Forcing them to enter their Postgres password, and showing them a warning message, means it will be less likely that this will happen.

scripts/RunningScripts.md
mattagape: This was addressed in 2fb55ad.

ga_dashboard/samples/grafana_users_list.csv
Llannelongue: I believe the original point stands: isn't it confusing to have an example file with columns that we don't expect users to include in their own version of the same file? I understand we don't have a full password pipeline yet, but then best to have a separate password file that we can clearly label as being for demonstration purposes only.

scripts/RunningScripts.md
Llannelongue: I don't think that's sufficiently explicit, there should be 2 distinct sections in the documentation: one for the first run of the code, and one for subsequent ones. Regardless of whether the same code is used, these are two separate processes.  Can be added to an issue and be resolved in the next PR.

scripts/RunningScripts.md
Llannelongue: The point of this PR to refactor the structure of the package and clarify how backend and frontend are connected, so it would be good to address it in this PR for clarity.

scripts/RunningScripts.md
Llannelongue: Will need to be clearer, but can be merged in as is for now.

Llannelongue:
(I've added comments on the recent commits)

scripts/backend/add_data_to_db.sh
Llannelongue: What is the scope of the packages? Ideally, we would want as much of the code in the package, making using it quite straight forward. But here, a lot of compulsory steps are in `/scripts/`. I appreciate it may have to be that way, just trying to understand what the reasons are for what is or isn't included in `ga_dashboard`.

mattagape: The software in `/scripts/` is, effectively, client code of the `ga_dashboard` package, and has examples of how the package should be used. As an example, someone might want to write `sacct` data to a PostgreSQL database, but use something other than Grafana to view their data.

ga_dashboard/samples/grafana_users_list.csv
mattagape: OK, I have now changed the workflow so that passwords are no longer stored in a CSV file. Instead, they are generated for a new user on the command line using a Python password generator library. You can set the options on this, e.g. minimum number of characters.

See beb85286fbc1e64f35f592bd7779d53093ea4d8e

mattagape:
This PR fixes #43 and #37.

scripts/RunningScripts.md
mattagape: New issue #46 created for this point.

scripts/RunningScripts.md
mattagape: OK, addressed in commit 19bcb5d.

scripts/backend/create_or_overwrite_database.sh
```diff
@@ -0,0 +1,61 @@
...
+db_name=$DEFAULT_DB_NAME
+db_host=$DEFAULT_DB_HOST
+db_port=$DEFAULT_DB_PORT
+db_user=$DEFAULT_DB_USER
+pg_version=$DEFAULT_POSTGRES_VERSION
+db_setup_script="$repo_root_dir/ga_dashboard/database/ga_db.sql"
+
+echo "\n*** Setting up empty Postgres database: ***\n"
+echo "\n** WARNING: this will delete any existing data in the database. **\n"
+echo "\n** Use CTRL-C to stop script. **\n"
```
Llannelongue: As I understand it, the only way to cancel is by doing ctrl+C very quick (do correct me if I'm wrong). Is so, that's risky. Could we replace this by an opt-in rather than opt-out? Deleting the database is pretty major, it would be good to ask users to confirm that it's really what they want to do before doing it.

scripts/RunningScripts.md
Llannelongue: Could we ask the users to confirm that deleting the existing database is what they really want? Something like "Press Yes to confirm deletion of the existing database"

scripts/backend/add_data_to_db.sh
Llannelongue: No I get that, my question is whether some of this shouls become part of the package. For example:
- `scripts/backend/add_users_to_database.py`
- or `scripts/backend/run_backend.py` (up to [here](https://github.com/GreenAlgorithms/GA4HPCdashboard/blob/b8193301508ad53a30153f0ece5d75c8166518a6/scripts/backend/run_backend.py#L93)

Most of these two files seem to be things that will be copy-pasted by all users (as I understand it, the arguments are required by the rest of the code anyway), so I'm just wondering whether it would make sense to move all that into the package. And, for example, IT teams could then simply call a function (either python function of command line depending on what's best):

```python
from ga_dashboard import add_users_to_db

add_users_to_db(users_list.csv)
```

Intuitively it sounds like an easier UX, but I appreciate there may be technical or UX reasons not to do it?

(The scripts in `scripts/frontend` don't seem to have much of this boilerplate code.)

scripts/RunningScripts.md
Llannelongue: Makes sense, so should everything (at least all the paths provided by users) be absolute path?

scripts/RunningScripts.md
Llannelongue: (duplicates other review thread)

scripts/RunningScripts.md
```diff
@@ -0,0 +1,138 @@
...
+* The directory given with the `--useOtherInfrastructureInfo` option must contain **both** the cluster information file (`cluster_info.yaml`) and the file listing the HPC users (`hpc_users_list.csv`); here, the `samples` subdirectory, which is part of the repo, is used.
+* In this example, we use the file `sacct_20250414.txt`, which you just generated. Sample `sacct`-output files are also in the `samples` subdirectory. 
+
+
+## Grafana set-up
+A number of things need to be done, assuming you have downloaded Grafana and are running its server. These are:
+* import the dashboard(s) you want to use into the Grafana server
+* add teams and users (and their information) to the Grafana server
+
+You must set the `--input_dir` argument, which specifies the directory containing your JSON files for the dashboards.
```
Llannelongue: Following up on https://github.com/GreenAlgorithms/GA4HPCdashboard/commit/abd223e5f36dac75ae69834f969c3073bb78f20c#r156236347: would it make sense to just have that flag sets as default to the dashboard json files we provide? (and the override is an optional flag). As I imagine IT teams will mostly not edit these?

Llannelongue:
Just going through the older comments to check if they're still relevant or require actions (I should have put them as reviews to avoid this, ah well):
- https://github.com/GreenAlgorithms/GA4HPCdashboard/commit/cbf447255072565baef8de2195e192eb5f9064ec#r156235594 
- https://github.com/GreenAlgorithms/GA4HPCdashboard/commit/7ad41342719d03c81b0d79bbfc2718788385108f#r156236207 -> this was the same comment I remade above re ctrl+c (at least I'm consistent :) )
- https://github.com/GreenAlgorithms/GA4HPCdashboard/commit/d71a30fe2d994ac93fba14f06dcda538e7dbf058#r156236239

scripts/backend/create_or_overwrite_database.sh
mattagape: The script paused indefinitely when the `postgres` password is requested. So it will stop and wait for user input. But I note your other comment, and I'll see if I can add an "Are you sure?" prompt.

scripts/RunningScripts.md
mattagape: Done, in commits b07ebf2 and 5b40284.

scripts/backend/create_or_overwrite_database.sh
mattagape: User now has to type YES to proceed (see commits b07ebf2 and 5b40284).

scripts/RunningScripts.md
mattagape: It's probably best to use the absolute path, but some scripts seem to allow relative paths on some options, as well.

mattagape: See e2e7a0b.

scripts/backend/add_data_to_db.sh
mattagape: Hmmm. The scripts are now in the following directories: `backend`, `database`, `frontend` and `end-to-end`, but I think the last one won't get used much now.

 It's certainly true that a number of arguments, for the `backend` and `database` scripts, are common to more than one script.

Also, if you look at the top of the script `add_users_to_database.py`, you'll see I'd already thought of something similar, by having a Database class. This would act as an abstraction wrapper around Postgres operations, to make interfaces cleaner and also be easier if a user wants to use a different sort of database.

The code will still need to get the options from the user somehow, e.g., `postgres` password. Some things could perhaps go in a config file, with a common parser class to extract arguments from this and then call the desired functions as required. But this would be a fair-sized chunk of work, ideally with its own branch and PR. I'll create an issue to remind us.

mattagape: See issue #48

scripts/RunningScripts.md
mattagape: In https://github.com/GreenAlgorithms/GA4HPCdashboard/commit/abd223e5f36dac75ae69834f969c3073bb78f20c#r156236347, I think you were proposing the opposite? Have you changed your mind?

ens-lgil: I think the default `--input_dir` was set to the demo directory and that's what @Llannelongue wanted to change ?

Llannelongue: Indeed :) In the first version, `--input_dir` was pointing at the demo by default, which has been fixed. But now it seems like it still needs to be specified by every user? And I imagine the vast majority of IT teams will just make it point to the JSON files we provide, so I'm wondering whether it would make more sense to just have it point there by default, making the flag optional and used only for people who want to override the dashboard JSON to use. Does that make sense?

scripts/backend/add_data_to_db.sh
Llannelongue: Ok let's revisit this for the beta and gamma versions.

scripts/RunningScripts.md
mattagape: OK, I'll set `--input_dir` to point to our dashboards directory by default, and the IT team can adjust as required.

mattagape: Done, in 9572eb7 !!

PR #54, opened 2025-08-06
Matt beta

mattagape:
Very simple PR; we can use one common file for HPC and Grafana users. I have created a sample of this. But we retain the option for there to be different HPC and Grafana user files, if so desired.


Llannelongue (changes requested):
I don't understand the approach taken here. As discussed, the point of having just one user file is to have just one file, and one input point for users, it's not to suggest that they may, if they want, use two files or use the same file once which appear to be unnecessarily confusing. 

Moreover, we now have 4 templates for user files. Any reason why we need to keep them all?
- `common_users_list.csv`
- `grafana_users_list.csv`
- `hpc_users_list.csv`
- `sample_user_list.csv`

And I'm not sure why it's called `samples/`, wouldn't it be clearer to be in `docs/template/` alongside the config file and the cluster info?

mattagape:
> I don't understand the approach taken here. As discussed, the point of having just one user file is to have just one file, and one input point for users, it's not to suggest that they may, if they want, use two files or use the same file once which appear to be unnecessarily confusing.
> 
> Moreover, we now have 4 templates for user files. Any reason why we need to keep them all?
> 
> * `common_users_list.csv`
> * `grafana_users_list.csv`
> * `hpc_users_list.csv`
> * `sample_user_list.csv`
> 

For simplicity, we can remove all the files you mention except for `sample_user_list.csv`. Although, I should point out, removing `grafana_users_list.csv` and `hpc_users_list.csv` means the end users will lose those examples.

I will need to update the documentation and scripts etc. to reflect the fact that we will now have only `sample_user_list.csv`. 

> And I'm not sure why it's called `samples/`, wouldn't it be clearer to be in `docs/template/` alongside the config file and the cluster info?

They are meant to be samples that the users can refer to when creating their own. The `docs` folder is just meant to contain documentation. But I can set up a `docs/templates/` directory and move the files there if you want. Again, I'll have to update the documentation etc.

mattagape:
OK @Llannelongue, I think I've now made the changes you requested.

Llannelongue:
> > And I'm not sure why it's called samples/, wouldn't it be clearer to be in docs/template/ alongside the config file and the cluster info?

> They are meant to be samples that the users can refer to when creating their own. The docs folder is just meant to contain documentation. But I can set up a docs/templates/ directory and move the files there if you want. Again, I'll have to update the documentation etc.

Looking at the new folder structure: not all files in this folder serve the same purpose, some are templates that we expect the IT teams to copy and edit, some are debugging testing files. Shouldn't they be in different folders?

Regardless, we shouldn't have a file called `df_agg_X_mockMultiUsers_1.csv` in a `docs/template/` folder, this will probably be mostly confusing for users.

README.md
```diff
@@ -15,8 +15,7 @@ Content
   * [Dashboard platform - Grafana](#dashboard-platform---grafana)
 * [Configuration files](#configuration-files)
   * [System configuration files](#system-configuration-files)
-  * [HPC users file](#hpc-users-file)
-  * [Grafana users file](#generate-a-grafana-users-file---csv-format)
+  * [HPC and Grafana users file](#list-of-users)
```
Llannelongue:
```suggestion
  * [List of users](#list-of-users)
```

README.md
```diff
@@ -29,17 +28,16 @@ Content
 Files required for the example instructions below (you can, of course, use your own):
```
Llannelongue: What does this section refer to? "example introduction"

README.md
```diff
@@ -102,67 +100,46 @@ As well as a list of users for the database, and another list for Grafana, the s
...
+* **Information about your HPC cluster**. Example: `docs/templates/cluster_info.yaml`.
 * **Fixed parameters file**. Example: `ga_dashboard/data/fixed_parameters.yaml`.
 
 
-### HPC users file
-You will need a file with details of your HPC users for whom you are obtaining `sacct` data (e.g. name, group).
-The HPC users file should be a comma-separated file combining these columns:
+### List of users
+
+You will need a file with details of your HPC users for whom you are obtaining `sacct` data (e.g. name, group), and the Grafana users. For simplicity, we are using the same file for both in this example.
```
Llannelongue: > For simplicity, we are using the same file for both in this example.

Not just in this example no? We are using a unique file throughout, there isn't a list of HPC user and a list of grafana users anymore, there is just a like of dashboard users.

README.md
```diff
@@ -102,67 +100,46 @@ As well as a list of users for the database, and another list for Grafana, the s
...
-
-In this example, we have the same list of users for both the HPC system and Grafana. But you may have some who are on one list only, e.g., a manager might want to have access to the Grafana dashboard, but not the HPC system.
 
+> [!WARNING]  
+> Do not store passwords in this file!
 
-An example you can use to try out the system is `ga_dashboard/samples/grafana_users_list.csv`
 
-The passwords generated by this process adhere to [Grafana password policy](https://grafana.com/docs/grafana/next/setup-grafana/configure-security/configure-authentication/grafana/#strong-password-policy), if you decide to enforce it.
+The passwords generated by the script which adds new users to Grafana (`import_users.py` or `setup_frontend.py`, both in `scripts/frontend/`) adhere to the [Grafana password policy](https://grafana.com/docs/grafana/next/setup-grafana/configure-security/configure-authentication/grafana/#strong-password-policy), if you decide to enforce it.
```
Llannelongue: Why are these scripts referenced here? We want to keep the number of user-facing scripts small

README.md
```diff
@@ -195,6 +172,8 @@ $ python scripts/run_green_algorithms_on_historical_logs.py --config <your_confi
 ```
 This will collect all the logs available by default (if no `startDay` / `endDay` are defined in the configuration file).
 
+Note: we have tested the software successfully with a `sacct`-output data file of more than one million entries. Our intention is to update the software so that it can safely handle much more than this.
```
Llannelongue: [This doesn't belong in this PR, in the future, it's good to keep the PRs to their main topic]

Good to have this. It would be good to also clarify what we expect of them at this stage: is it to only run it on up to a million rows? If so what can they expect in terms of run time and memory pressure?

scripts/frontend/x.py
```diff
@@ -0,0 +1,4 @@
+import os
```
Llannelongue: What is this file for?

scripts/gamma-stress-config.txt
```diff
@@ -0,0 +1,108 @@
+# Sample config file
```
Llannelongue: Based on the name, I don't think this belongs in this PR?

scripts/my_config.txt
```diff
@@ -0,0 +1,80 @@
+# Config file templates
```
Llannelongue: What is this file?

Llannelongue (changes requested):

scripts/frontend/x.py
mattagape: Oops, added by mistake. Removed in dff361c.

scripts/gamma-stress-config.txt
mattagape: Oops, added by mistake. Removed in dff361c.

scripts/my_config.txt
mattagape: Oops, added by mistake. Removed in dff361c.

README.md
mattagape: Indeed, see 63e26db and 0d2164c.

README.md
mattagape: Yes, I am going through everywhere to make this clear, e.g. 63e26db and 909adbd.

README.md
mattagape: Fixed in 9cb3780.

README.md
mattagape: I'll add this once I've run a proper stress test with quantitative results.

README.md
mattagape: I've removed the details of the specific scripts. I'll change the rest of the wording before the `beta-passwords` branch is merged.

mattagape:
> > > And I'm not sure why it's called samples/, wouldn't it be clearer to be in docs/template/ alongside the config file and the cluster info?
> 
> > They are meant to be samples that the users can refer to when creating their own. The docs folder is just meant to contain documentation. But I can set up a docs/templates/ directory and move the files there if you want. Again, I'll have to update the documentation etc.
> 
> Looking at the new folder structure: not all files in this folder serve the same purpose, some are templates that we expect the IT teams to copy and edit, some are debugging testing files. Shouldn't they be in different folders?
> 
> Regardless, we shouldn't have a file called `df_agg_X_mockMultiUsers_1.csv` in a `docs/template/` folder, this will probably be mostly confusing for users.

I've now renamed those files. See 2154997 and 732d24b.

README.md
```diff
@@ -59,143 +57,123 @@ $ conda env list
...
+* **Information about your HPC cluster**. Example: `docs/templates/cluster_info.yaml`.
 * **Fixed parameters file**. Example: `ga_dashboard/data/fixed_parameters.yaml`.
 
 
-### HPC users file
-You will need a file with details of your HPC users for whom you are obtaining `sacct` data (e.g. name, group).
-The HPC users file should be a comma-separated file combining these columns:
+### List of users
+
+You will need a file with details of your Dashboard users. For each, you will obtain their `sacct` data, and add them to the Postgres database and the Grafana instance.
```
Llannelongue:
```sugggestion
 This is a file with details of your Dashboard users, to collate their HPC use and create a Grafana account for them.
```

README.md
```diff
@@ -59,143 +57,123 @@ $ conda env list
...
+| uid_1 | 11111 | John Smith | user1@example.com | group 1 | Dept 3     | *0IK^I^&UpO$2aX |
+| uid_2 | 22222 | Sarah Jones | user2@example.com   | group 1 | Dept 3     | yGg=kA-6v**7BS) |
+| ...   | ...   | ...           | ...     | ...  | ...      | ... |
 
 
-An example you can use to try out the system is `ga_dashboard/samples/grafana_users_list.csv`
+The passwords in the above example adhere to the [Grafana password policy](https://grafana.com/docs/grafana/next/setup-grafana/configure-security/configure-authentication/grafana/#strong-password-policy), should you decide to enforce it.
 
-The passwords generated by this process adhere to [Grafana password policy](https://grafana.com/docs/grafana/next/setup-grafana/configure-security/configure-authentication/grafana/#strong-password-policy), if you decide to enforce it.
+Make sure the passwords don't contain a comma character (`','`), otherwise this will affect the CSV file parsing.
```
Llannelongue, line 142: We should make a note to allow IT teams to pass a csv file with another separator to avoid this. Not necessarily for the beta.

README.md
```diff
@@ -212,14 +190,14 @@ The 2 scripts proceed to:
 
 The (anonymised) examples of files you can use with `useCustomLogs` are:
```
Llannelongue: Why do we discuss this? It should be attached to a clear usecase.

README.md
```diff
@@ -246,7 +224,7 @@ $ cd /.../grafana/
 $ ./bin/grafana server
 ```
 
-Depending on your system, you may not be able to do this. For example, on Linux, you will need to use these steps to run the Grafana server after installation:
+Depending on your system, you may not be able to do this. For example, on Linux, you might need to use these steps to run the Grafana server after installation:
```
Llannelongue: As discussed previously, UNIX/Linux should be the default in the documentation

docs/templates/about_test_files.md
```diff
@@ -0,0 +1,22 @@
+Explanation of test files
+--------------------
+
+These are anonymised test files to use for debugging and development. Their main function is to bypass parts of the code, for example to be able to run it offline, or to simulate multiple users without admin access. 
+
+The two files that are most relevant are:
+- To run the code offline on a single user, use `--useCustomLogs sacct_output_single_user.txt`
+- To run the code on multiple users (either offline or without admin rights), use `--use_mock_agg_data`, which will use `extracted_multi_users.csv`.
+- To run only the database generation (and possibly frontend), bypassing the backend, use `aggregated_multi_users.csv`.
```
Llannelongue: the command flag to use is missing here

docs/templates/about_test_files.md
```diff
@@ -0,0 +1,22 @@
+Explanation of test files
```
Llannelongue: Good to have this doc, but same as the files discussed in here, it doesn't belong in `templates/`

README.md
mattagape: Changed to this in 1fd3578.

README.md
mattagape: Noted in issue #58 .

README.md
mattagape: Updated text in b5216f1.

README.md
mattagape: Indeed. Clarified in 1d0caa2.

docs/templates/about_test_files.md
mattagape: I'll move them, in a separate PR, to `tests/testdata/`. And update documentation etc. in that same PR.

docs/templates/about_test_files.md
mattagape: At the moment this case is handled by the special script `scripts/database/import_mockup_aggregate.py`. I can try and add this as an option to Laurent's new Python scripts in a separate PR.

I've created issue #60

docs/templates/about_test_files.md
mattagape: See issue #61

Llannelongue: As discussed, this should be included in this PR. Let me know when done.

README.md
```diff
@@ -210,16 +187,18 @@ The 2 scripts proceed to:
...
-* `ga_dashboard/samples/sacct_output_single_user.txt`
+Examples (anonymised) in the file format required to use this flag are:
+
+* `tests/testdata/sacct_output_single_user.txt`
   > Example of output generated, for one user, by the `sacct` command on the HPC system.
-* `ga_dashboard/samples/sacct_output_multi_user.txt`
+* `tests/testdata/sacct_output_multi_user.txt`
   > Same example as above, but for multiple users.
 
 The backend part of the software will aggregate the data into one row per user per day, enrich it (add carbon footprint data), and then write this to the database.
```
Llannelongue:
```suggestion
> [!NOTE]
> The scripts above have the same admin user run the SLURM `sacct` command to download all the logs, process these logs, and add the processed data to the Postgres database. In some cases, it may not be suitable (e.g. if you want to run the `sacct` command on one machine, transfer the data to another one hosting the database and the two can't communicate directly. We haven't developed this alternative pipeline in the beta version quite yet, but if this is your case, do get in touch, we can walk you through separating the two parts of the code. 
```

README.md
```diff
@@ -261,53 +240,9 @@ In the former case above, you can just CTRL-C the server. In the latter, you mig
...
-  --admin_login admin --admin_password <adm_password> \
-  --db_name ga_db --db_user <db_user_name> --db_password <db_user_password> --pg_version <postgresql_version> \
-  --input_file <path_to_users_csv_file>
-```
-
-Using the default options, and default password for Grafana admin:
-```
-$ python scripts/frontend/setup_frontend.py --admin_password admin --db_password <password>
-```
+The `systemctl` command might be elsewhere on a Linux system, e.g., `/usr/bin/systemctl`.
```
Llannelongue, line 235: I'm not sure what this section is doing, if not needed, can be deleted

README.md
mattagape: Done in b365bb2

docs/templates/about_test_files.md
mattagape: Done in commits [f104900](https://github.com/GreenAlgorithms/GA4HPCdashboard/pull/54/commits/f10490044cfc026fb0710e4f4f983df9baa8d8af) et seq.
Fixes #61

README.md
mattagape: You mean the line:
```
The `systemctl` command might be elsewhere on a Linux system, e.g., `/usr/bin/systemctl`.
```
That's to help users, whose Linux might be different. I'll leave it in.

PR #62, opened 2025-08-18
Beta behaviour

mattagape:
Mainly to ensure Grafana and Postgres both update users correctly.


Llannelongue:
Quick question on this PR: are the changes to the following files related to adding users across the DB and Grafana?
- `setup_environment.sh`
- and the files under `tests/`

mattagape:
> Quick question on this PR: are the changes to the following files related to adding users across the DB and Grafana?
> 
> * `setup_environment.sh`
> * and the files under `tests/`

Good catch. I'm afraid not. The files under `/tests` had to be changed so that the unit tests would work after I removed the duplicate files from there. The environment script I added a single extra line as it kept failing before I did so.

Llannelongue:
Ok, then can you please amend the PR to only adress this PR's topic? (or delete it and start a new PR)

Llannelongue (changes requested):
Remove changes not related to the PR's topic

mattagape:
Reverted the changes so they can be done on another PR.

Llannelongue:
Closing at this stage since the code doesn't run as is.

PR #69, opened 2025-08-27
Beta install

mattagape:
Add timing information to `scripts/run_green_algorithms_on_historical_logs.py` to help with stress tests. And update/correct the README. A couple of minor changes from Laurent.


README.md
```diff
@@ -55,13 +55,24 @@ $ conda env list
...
+$ poetry install
+```
+based on which tool (`pip` or `poetry`) you prefer to use.
+
+To install the `ga_dashboard` software package specifically, type:
 ```
 $ python -m pip install .
 ```
-(Note the period character at the end). This should install the `ga_dashboard` package on your local machine. if you want to be able to 
+(Note the period character at the end). This should install the `ga_dashboard` package on your local machine. If you want to be able to
```
Llannelongue, line 75: >  If you want to be able to edit it and still use it, use the `-e` option:

Not sure what that refers to?

README.md
```diff
@@ -180,7 +191,9 @@ $ python scripts/run_green_algorithms_on_historical_logs.py --config <your_confi
 ```
 This will collect all the logs available by default (if no `startDay` / `endDay` are defined in the configuration file).
 
-**Note:** we have tested the software successfully with a `sacct`-output data file of more than one million entries. Our intention is to update the software so that it can safely handle much more than this.
+**Note:** we have tested the software successfully with a `sacct`-output data file of more than one million entries (actually 1,072,406 lines of data, including the header, which must be present). On a Mac, running the `run_green_algorithms_on_historical_logs.py` script on this data took 11m 42s, and the peak memory usage (measured with `mprof`) was around 4.7 GB. (Another run showed that the overhead due to `mprof` only made a small difference). This was an optimised situation; `sacct` had been run previously (and generated the file of results to use), and the files and Python scripts, Postgres database, and Grafana server were all running on the same local machine.
```
Llannelongue:
```suggestion
**Note:** we have tested the software successfully on up to 1M jobs' log files (i.e. the `sacct` command returns one million entries). On a Mac, running the `scripts/run_green_algorithms_on_historical_logs.py` script on this data took 11m 42s **without the `sacct` runtime**, and the peak memory usage (measured with `mprof`) was around 4.7 GB.  This was an optimised situation; `sacct` had been run previously (and generated the file of results to use), and the files and Python scripts, Postgres database, and Grafana server were all running on the same local machine.

For the beta testing, you may want to restrict the dates to a few months to not have more than a few million jobs returned (until we have tested scalability in more details).
```

README.md
mattagape: It's in the line after that:
```
$ python -m pip install -e .
```

Llannelongue: No I know, but I don't understand what "If you want to be able to edit it and still use it" means?

mattagape: It means: "Install it, but allow user to edit and change it if necessary."

Llannelongue: Ah ok. In the documentation, if we mention it, we should make clear what use case this refers to, so users know what to do with this information.

mattagape: I'll add it to the README.

mattagape: OK, done in d8ad353.

PR #71, opened 2025-08-28
Remove regex warning and remove the 'outFile' attribute from the mand…

ens-lgil:
…atory parameters


mattagape (changes requested):
I think we need to make sure that the user **cannot choose both** `outFile` and `useCustomLogs` options.

mattagape:
That's fine. Please go ahead and merge.

mattagape (approved):

PR #83, opened 2025-09-25
Prepare code for new demo branch

ens-lgil:
- Update config demo file
- Add optional password parameters in the installation/setup scripts
- Update demo shell script: use the new python scripts and use the `config__demo.txt` file instead of a list of variables.


Llannelongue:
Just so I understand, what is `scripts/end-to-end/demo.sh` file for?

ens-lgil:
It seems to be a wrapper to run the install scripts with the demo configuration

PR #134, opened 2026-02-13
Db sacct refactoring

Navirah:
Issue #135 

The main changes done as a part of this PR includes
1. Database refactoring
2. Sacct refactoring

Both of the above now exist as a **Service** (ga_dashboard>backend>services). A **Service** is a discrete unit of functionality that exposes specific capabilities to other parts of a system. Example, in the current scenario it is a DatabaseService class that encapsulates all database related functions.

Note: As part of a larger refactoring, many of the current functionality can be redesigned as a service. Like Workload_manager etc.

___________________

I've also tried to collect all helper/util functions under a the **helpers** function in ga_dashboard>backend. Going ahead. we can add such generic functions here.

____________________

In slurm.py, I've made **convert2dataframe()** a static method since it seemed general enough and need to use the instance of the class i.e. self.


ga_dashboard/backend/workload_manager/wm_helpers.py
```diff
@@ -0,0 +1,221 @@
+import datetime
+import pandas as pd
+import numpy as np
+
+class WMHelpers:
+
```
ens-lgil: Would be nice to add some description about the class, as you did for the services classes.

scripts/database/import_mockup_aggregate.py
```diff
@@ -105,110 +107,89 @@ def main():
...
-    # Close communication with the database
-    cur.close()
-    conn.close()
-
-    print ("Script completed successfully.")
+            user_rows.append(user_data_row)
+
+    # Use DatabaseService for bulk inserts
+    with DatabaseService(db_settings) as db:
+        if not db._conn or db._conn.closed:
```
ens-lgil: Would it be better to not call these internal functions outside the class, and create a `DatabaseService.is_conn_ok()` function for instance to test it and return a boolean

ens-lgil (commented):
Look really nice, especially with the new services modules!

ga_dashboard/backend/workload_manager/wm_helpers.py
Navirah: For sure! I've added it in the new commit. Also renamed this class for better understandability.

ga_dashboard/backend/helpers/logging_utils.py
```diff
@@ -0,0 +1,23 @@
+
+import datetime
+import os
+from ga_dashboard.backend.workload_manager.slurm import WorkloadManager
+
+def report_bugs(config_data, WM: WorkloadManager) -> None:
```
Llannelongue: This is a legacy of GA4HPC: because with GA4HPC I couldn't see other people's log, debugging was difficult. This code gives the option to export the logs somewhere where I could see them to test the code on these and see what doesn't work. The logic may need adapted to the dashboard? (maybe not, just asking here!)

ga_dashboard/backend/services/database_service.py
```diff
@@ -0,0 +1,191 @@
+import logging
+import psycopg
```
Llannelongue, line 6: Could you just explain the structure with `services/database.py` in the `backend` and some other database things in the `database` folder?

ga_dashboard/database/table_col_definitions.py
```diff
@@ -0,0 +1,19 @@
+# ------------------------------------------------------------------
+# Column definitions (add new tables here)
```
Llannelongue, line 2: Would it be worth having a comment like that at the top of most python files to clarify what they do? I'm mindful it's now quite a few of them!

ga_dashboard/backend/workload_manager/WorkloadDataHandler.py
```diff
@@ -0,0 +1,225 @@
+import datetime
+import pandas as pd
+import numpy as np
+
+class WorkloadDataHandler:
+    """
+    A utility class for processing and analyzing workload data from HPC cluster job schedulers. 
+    Handles data transformation, parsing, cleaning, and metric calculations to support
```
Llannelongue: I think this is very specific to SLURM (the name of the fields, the format etc.), so it would be good to make it clear now that it's a separate file. Maybe worth having a `slurm` subfolder in `workload_manager`?

Llannelongue (commented):
Looks a lot neater :) I've added some comments/questions

ga_dashboard/backend/workload_manager/WorkloadDataHandler.py
ens-lgil: Yes, we might need some sorts of adaptor to "map" the columns of the different workload manager logs to our database fields/tables (which are currently pretty much the same as the SLURM fields).

ga_dashboard/backend/helpers/logging_utils.py
Navirah: I think it's useful but on second thoughts I feel like the function should be renamed. It kind of insinuates that we're reporting bugs but instead we're just saving WM.logs_raw for debugging.

ga_dashboard/backend/services/database_service.py
Navirah: I realized that database related things now lie in two different places and 'database.py' makes it ambiguous, therefore I had renamed the file to database_service.py - which includes read/write functions. 

The database related things in ga_dashboard>database are more schema related. Maybe it can be renamed to database_schema.

ga_dashboard/backend/workload_manager/WorkloadDataHandler.py
Navirah: Got your point. For now I think creating a slurm folder under workload_manager should do. As part of a bigger refactoring we can look at adding an adapter to accommodate other types of workload managers. Will create a separate issue for this.

ga_dashboard/backend/helpers/logging_utils.py
Navirah: I've changed the name of this function and placed it in utils.py in latest commit.

ga_dashboard/database/table_col_definitions.py
Navirah: Added short descriptions over python files (most of them that could use it) in the latest commit

ga_dashboard/backend/workload_manager/WorkloadDataHandler.py
Navirah: Resolved in latest commit. Issue created to make it more flexible in the future.

ga_dashboard/backend/services/validation_service.py
```diff
@@ -0,0 +1,97 @@
...
+# Service constaining the main backend validation logic (any new validation logic should be added here)
+# ------------------------------------------------------------------
+
+import datetime
+
+class validate_args:
+    """
+    Class used to validate all the arguments provided.
+    """
+    # TODO add validation
```
Llannelongue: Minor point: are these 2 TODOs still relevant? If not, feel free to delete

ga_dashboard/backend/workload_manager/slurm.py
Llannelongue, line 1: Is this the main file for the SLURM part of the backend? `slurm_workload_manager.py` sounds redundant as a filename (SLURM is indeed a workload manager). If this is the main file, could the content be moved to `__init__.py` instead to keep things simple? (just a suggestion, do feel free to disagree!)

ga_dashboard/backend/data_sql_import.py
```diff
@@ -1,3 +1,7 @@
+# ------------------------------------------------------------------
+# Bridge betweem main_backend in ga_tools.py and DatabaseService
```
Llannelongue, line 2: It's probably me, but I don't have a clear idea of what this means :)

Llannelongue (approved):
I left some minor comments/suggestions, but looks good to merge in!

ga_dashboard/backend/services/database_service.py
Llannelongue: Ok makes thanks, thanks!

ga_dashboard/backend/helpers/logging_utils.py
Llannelongue: Sounds good

ga_dashboard/backend/services/validation_service.py
Navirah: @ens-lgil do you know know what these TODOs were for?

ga_dashboard/backend/data_sql_import.py
Navirah: Because of the way the code is currently written, data_sql_import becomes a place where the data is processed just before storing it in the Database. So essentially a bridge step between GA processing (cleaning of logs and calculating emissions) and Database operations. I think we're mainly performing column mapping and datatype conversions to make it consistent with the database tables. Hope that makes sense. There's definitely scope of refactoring here. Will take it up as part of later Issues.

ga_dashboard/backend/workload_manager/slurm.py
Navirah: This makes sense. I've simplified it a bit (basically reverted to the old slurm.py but renamed the classes to make more sense). 
Since the file had substantial code, I thought against putting it in __init__.py because it might make the code harder to navigate. 

Something for later: the parent and child class relationship in slurm.py doesn't make much sense since SlurmManager does not have an 'is-a' relationship with SlurmBase, it just makes use of its functions. Composition would be more appropriate here. This would require some light refactoring.

ga_dashboard/backend/services/validation_service.py
ens-lgil: I think Matt forgot to remove them when he was working on this function.
However we might add more code to this function in the future.

Navirah: I've removed these comments in the latest commit. As we discussed yesterday, any new validation to be added can be done in validation_service file going forward.

Issue #133, opened 2026-02-10
Handling UNFINISHED jobs - Reverted


Navirah:
_The problem at hand:_

Since we are using a starttime and endtime to fetch the jobs for the day, lets say 08-02-2026T00:00:00 to 08-02-2026T23:59:59, It will pick jobs that started in this interval. If the job started in this interval but extended to the next one lets say till 11am on 09-02-2026, the remaining would not be picked up when 09-02-2026 is processed. If the Dashboard job runs at, let's say, 4am, it will consider 4 hours of work done on 09-02-2026, but not after. 

_Scope:_
- Create a separate table in the database to store Jobs from SLURM that have not yet completed their run , i.e. their status is RUNNING or end time is not specified (need to check the field). 
- The carbon footprint of these RUNNING jobs will not be accounted for in the current Dashboard Job.
- During each Dashboard Job run, we must fetch the stored RUNNING Jobs from the DB and run a SLURM command to fetch their current state using JobID as parameter. 
- If these fetched jobs are finished, their corresponding carbon footprint must be accounted for in the current Dashboard Job and then the entry should be removed from the DB. If not, the unfinished job must remain in the DB to be fetched in the next Dashboard Job run.

Navirah:
Reverted due to redundancy. 
Check out #159

PR #145, opened 2026-03-06
Handling Unfinished Jobs

Navirah:
Issue #133 
(process notes with help from ClaudeAI)

## Unfinished Jobs Handling

Jobs still in `PENDING`, `RUNNING`, `SUSPENDED`, `PREEMPTED`, or `UNKNOWN` state at pipeline runtime can't be fully accounted for. Following steps capture them and re-processes them once they finish.

### Lifecycle across two runs

**Run N**
1. Pull Slurm logs
2. Detect unfinished jobs — checks `End` date first (missing or `"Unknown"`); falls back to `State` only if `End` column is absent
3. Save them to the `ga_unfinished_jobs` DB table

**Run N+1**
1. Fetch previously stored job IDs from `ga_unfinished_jobs`
2. Re-query `sacct` by Job ID to check for completion
3. Re-inject any now-finished jobs into the main dataframe
4. Enrich → Summarise → Insert to DB
5. **Only then** delete resolved jobs from `ga_unfinished_jobs` (prevents data loss if insert fails)

### Relevant files
- `unfinished_jobs_service.py` — DB interactions and `sacct` re-querying
- `ga_tools.py` → `extract_data()` — orchestrates detection, recovery, re-injection
- `ga_tools.py` → `process_and_store()` — triggers safe deletion post-insert

### State reference

| | States |
|---|---|
| **Finished** | `BOOT_FAIL`, `CANCELLED`, `COMPLETED`, `DEADLINE`, `FAILED`, `NODE_FAIL`, `OUT_OF_MEMORY`, `TIMEOUT` |
| **Unfinished** | `PENDING`, `RUNNING`, `SUSPENDED`, `UNKNOWN`, `PREEMPTED` |

### DIFF
--- a/ga_dashboard/backend/ga_tools.py
+++ b/ga_dashboard/backend/ga_tools.py
@@ -126,7 +126,7 @@ def extract_data(config_data: dict, has_slurmAdmin: bool, cluster_info) -> pd.Da
 
 
     ### Turn usage logs into DataFrame
-    WM.convert2dataframe()
+    WM.raw_logs_to_df()
 
     ###### Running jobs can be filtered and saved into the DB here. 
     ###### Subsquently they can be removed from WM logs before proceeding with cleaning of logs.
--- a/ga_dashboard/backend/ga_tools.py
+++ b/ga_dashboard/backend/ga_tools.py
@@ -131,8 +131,9 @@ def extract_data(config_data: dict, has_slurmAdmin: bool, cluster_info) -> pd.Da
   
     WM.filter_and_store_unfinished_jobs(unfin_jobs_service) # Filter unifinshed jobs from WM logs and store them in DB
 
-    WM.concat_logs_df(finished_jobs_df) #concat finished jobs (previously unfinished) with rest of the logs
-    ### TODO Delete the jobs that have finished from DB
+    if finished_jobs_df is not None:
+        WM.concat_logs_df(finished_jobs_df) #concat finished jobs (previously unfinished) with rest of the logs
+
     # And clean
     WM.clean_logs_df()
     # Check if there are any jobs during the period from this directory and with these jobIDs
@@ -145,7 +146,7 @@ def extract_data(config_data: dict, has_slurmAdmin: bool, cluster_info) -> pd.Da
 
     # WM.df_agg_X.to_pickle("testdata/df_agg_X_1.pkl") # DEBUGONLY used to test different steps offline
 
-    return WM.df_agg_X
+    return WM.df_agg_X, unfin_jobs_service
 
 
 def enrich_data(df: pd.DataFrame, fixed_params: dict, users_df: pd.DataFrame, GA: GA_tools) -> pd.DataFrame:
@@ -330,7 +331,7 @@ def main_backend(config_data: dict):
 
     GA = GA_tools(cluster_info, fixed_params)
 
-    df = extract_data(config_data, has_slurmAdmin, cluster_info=cluster_info)
+    df, finished_jobids = extract_data(config_data, has_slurmAdmin, cluster_info=cluster_info)
     df2 = enrich_data(df, fixed_params=fixed_params, users_df=users_df, GA=GA)
     summary_stats = summarise_data(df2) # TODO export and save df_userdaily regularly (as a more manageable database than all individual jobs?)
 
@@ -341,12 +342,11 @@ def main_backend(config_data: dict):
     # except Exception:
     #     dict_keys = args._asdict().keys() # This is when using the debugging namedtuples TODO this a bit messy, should be cleaned up
     if 'db_name' in dict_keys:
-        process_and_store(summary_stats, config_data)
-
+        process_and_store(summary_stats, config_data, finished_jobids)
     return summary_stats
 
 
-def process_and_store(summary_stats:dict,config_data:dict) -> None:
+def process_and_store(summary_stats:dict,config_data:dict, unfin_jobs_service:UnfinishedJobsService) -> None:
     # Import aggregated data into a database
     db_params = DBSettings(
         db_name=config_data['db_name'],
@@ -359,4 +359,13 @@ def process_and_store(summary_stats:dict,config_data:dict) -> None:
                 summary_stats,
                 db_params
             )
-    data2db.insert_data_into_db()
+    try: 
+        data2db.insert_data_into_db()
+    except Exception as e:
+        print(f"Error occurred while inserting data into database: {e}")
+        return
+    
+    #only delete the finished jobs from ga_unfinished_jobs table after the new data has been successfully inserted
+    unfin_jobs_service.delete_resolved_unfinished_jobs()
+    
+        
--- a/ga_dashboard/backend/ga_tools.py
+++ b/ga_dashboard/backend/ga_tools.py
@@ -126,11 +126,13 @@ def extract_data(config_data: dict, has_slurmAdmin: bool, cluster_info) -> pd.Da
     ### Turn usage logs into DataFrame
     WM.raw_logs_to_df()
 
-    finished_jobs_df = unfin_jobs_service.sync_unfinished_jobs() #Adding finished jobs to WM before cleaning and processing 
-    WM.concat_logs_df(finished_jobs_df) 
-    ###### Running jobs can be filtered and saved into the DB here. 
-    ###### Subsquently they can be removed from WM logs before proceeding with cleaning of logs.
+    # get unfinished jobs from db and pick those that have now finished using sacct command by job id
+    finished_jobs_df = unfin_jobs_service.sync_unfinished_jobs() 
+  
+    WM.filter_and_store_unfinished_jobs(unfin_jobs_service) # Filter unifinshed jobs from WM logs and store them in DB
 
+    WM.concat_logs_df(finished_jobs_df) #concat finished jobs (previously unfinished) with rest of the logs
+    ### TODO Delete the jobs that have finished from DB
     # And clean
     WM.clean_logs_df()
     # Check if there are any jobs during the period from this directory and with these jobIDs
@@ -228,7 +230,7 @@ def summarise_data(df: pd.DataFrame) -> dict:
             timeseries = data.groupby(agg_names2).agg(**agg_functions_from_raw)
         else:
             timeseries = data.groupby(agg_names2).agg(**agg_functions_further)
-
+ 
         timeseries.reset_index(inplace=True, drop=(agg_names is None))
         timeseries['success_rate'] = timeseries.n_success / timeseries.n_jobs
         timeseries['failure_rate'] = 1 - timeseries.success_rate
@@ -236,8 +238,7 @@ def summarise_data(df: pd.DataFrame) -> dict:
 
         return timeseries
 
-    df['SubmitDate'] = df.SubmitDatetimeX.dt.date  # TODO do it with real start time rather than submit day
-
+    df['SubmitDate'] = df.SubmitDatetimeX.dt.date  # TODO do it with real start time rather than submit date
     has_slurmAdmin = True # get_slurmAdmin(args) # We only assume we have admin access now
 
     if has_slurmAdmin:
@@ -262,6 +263,7 @@ def summarise_data(df: pd.DataFrame) -> dict:
 
         df_overallStats = agg_jobs(df_daily)
         dict_overallStats = df_overallStats.iloc[0, :].to_dict()
+
 
         ## And put everything in a dict
         output = {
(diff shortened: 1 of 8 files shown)


ga_dashboard/backend/helpers/utils.py
```diff
@@ -101,3 +102,28 @@ def save_slurm_logs(config_data, WM) -> None: # DEBUGONLY
         print(f"\nSLURM statistics saved for inspection: {log_path}\n")
 
 
+def convert2dataframe(df_raw, types=None, delimiter="|"):
```
ens-lgil: Could you add the type of `df_raw` and `types` ?

ga_dashboard/backend/services/unfinished_jobs_service.py
```diff
@@ -0,0 +1,117 @@
...
+## States from SLURM documentation: https://slurm.schedmd.com/job_state_codes.html (as of 10 Feb 2026)
+
+SLURM_STATES = ['BOOT_FAIL','CANCELLED','COMPLETED','DEADLINE','FAILED','NODE_FAIL','OUT_OF_MEMORY','PENDING','PREEMPTED','RUNNING','SUSPENDED','TIMEOUT','UNKNOWN']
+FINISHED_STATES = ['BOOT_FAIL','CANCELLED','COMPLETED','DEADLINE','FAILED','NODE_FAIL','OUT_OF_MEMORY','TIMEOUT']
+UNFINISHED_STATES = ['PENDING','RUNNING','SUSPENDED','UNKNOWN','PREEMPTED']
+
+class UnfinishedJobsService:
+    def __init__(self, config_data):
+        self.config_data = config_data
+        self.db_params = DBSettings(
```
ens-lgil: Just wondering if we could write a parent class to build the DBSettings so we would avoid to write each DB settings lines in this class and in [DatabaseService](https://github.com/Cambridge-Sustainable-Computing-Lab/Green-Algorithms-HPCdashboard/blob/main/ga_dashboard/backend/services/database_service.py#L29).

ga_dashboard/backend/workload_manager/slurm.py
```diff
@@ -466,3 +448,51 @@ def clean_logs_df(self):
                 self.df_agg = self.df_agg.loc[self.df_agg.Account_ == self.config_data['filterAccount']]
 
         self.df_agg_X = self.df_agg[[x for x in self.df_agg.columns if x[-1] == 'X']]
+
+    def concat_logs_df(self, new_logs_df):
```
ens-lgil: Add `new_logs_df` type:
```
def concat_logs_df(self, new_logs_df: pd.DataFrame]:
```

ga_dashboard/backend/services/unfinished_jobs_service.py
```diff
@@ -0,0 +1,117 @@
...
+from ga_dashboard.database.table_col_definitions import UNFINISHED_JOBS_COLUMNS
+
+## States from SLURM documentation: https://slurm.schedmd.com/job_state_codes.html (as of 10 Feb 2026)
+
+SLURM_STATES = ['BOOT_FAIL','CANCELLED','COMPLETED','DEADLINE','FAILED','NODE_FAIL','OUT_OF_MEMORY','PENDING','PREEMPTED','RUNNING','SUSPENDED','TIMEOUT','UNKNOWN']
+FINISHED_STATES = ['BOOT_FAIL','CANCELLED','COMPLETED','DEADLINE','FAILED','NODE_FAIL','OUT_OF_MEMORY','TIMEOUT']
+UNFINISHED_STATES = ['PENDING','RUNNING','SUSPENDED','UNKNOWN','PREEMPTED']
+
+class UnfinishedJobsService:
+    def __init__(self, config_data):
```
ens-lgil: Add `config_data` type:
```
def __init__(self, config_data: dict):
```

ga_dashboard/backend/workload_manager/slurm.py
```diff
@@ -466,3 +448,51 @@ def clean_logs_df(self):
...
+        - oldest Start date
+        - Submit date (should be the same across rows)
+        - State that is unfinished (if any subprocess is unfinished, the job is unfinished)
+        '''
+        if self.unfinished_jobs is None:
+            raise ValueError("unfinished_jobs is not initialised. Run filter_and_store_unfinished_jobs() first.")
+
+        df = self.unfinished_jobs.copy()
+
+        def pick_unfinished_state(states):
```
ens-lgil: Type for `states`

ga_dashboard/database/table_col_definitions.py
```diff
@@ -16,4 +16,4 @@
 
 GA_USER_COLUMNS = ['user_name', 'uid', 'name', 'group_name', 'department', 'updated']
 
-RUNNING_JOBS_COLUMNS = ['job_id', 'submite_date']
\ No newline at end of file
+UNFINISHED_JOBS_COLUMNS = ['job_id', "user_name",'submitdate', 'startdate','job_state']
```
ens-lgil: Use single quotes and spance for consistency:
```
UNFINISHED_JOBS_COLUMNS = ['job_id', 'user_name', 'submitdate', 'startdate', 'job_state']
```

ga_dashboard/backend/services/unfinished_jobs_service.py
```diff
@@ -0,0 +1,117 @@
...
+                user=self.config_data['db_user'],
+                password=self.config_data['db_password'],
+                host=self.config_data['db_host'],
+                port=self.config_data['db_port']
+                )
+
+    def filter_unfinished_jobs(self, logs_df: pd.DataFrame) -> pd.DataFrame:
+        '''
+        
+        Filter unfinished jobs from the logs dataframe using the 'State' column (if 'End' column is not available) or the 'End' column (if available).
```
Llannelongue, line 29: What could cause the "end" column to not be available?

ga_dashboard/backend/services/unfinished_jobs_service.py
```diff
@@ -0,0 +1,117 @@
...
+        return logs_df[mask].copy(), logs_df[~mask].copy() # Return both unfinished and finished jobs for further processing
+    
+    def filter_finished_jobs(self, logs_df: pd.DataFrame) -> pd.DataFrame:
+        '''
+        Filter finished jobs from the logs dataframe using the 'State' column (if 'End' column is not available) or the 'End' column (if available).
+        '''
+        if 'End' in logs_df.columns:
+            mask = logs_df['End'].notna() & (logs_df['End'] != "Unknown")
+        else:
+            mask = logs_df['State'].isin(FINISHED_STATES)
```
Llannelongue: A risk of using `.isin(FINISHED_STATES)` instead of `not .isin(UNFINISHED_STATES)` is that if for some reason our list of state is not exhaudtive, we could be missing out on some jobs.

ga_dashboard/backend/services/unfinished_jobs_service.py
```diff
@@ -0,0 +1,117 @@
...
+        return logs_df[mask].copy()
+    
+    def sync_unfinished_jobs(self) -> pd.DataFrame:
+        '''
+        Get the list of unfinished jobs from the previous extraction (if any) to check if they are still unfinished or if they have finished in the meantime.
+        Pull the jobs using sacct command on the basis of Job IDs
+        '''
+        use_mock = True if 'use_mock' in self.config_data.keys() and self.config_data['use_mock'] else False
+        if use_mock:
+            prev_unfinished_jobs_df = pd.read_csv(self.config_data['use_mock'])
```
Llannelongue: is `config_data['use_mock']` only unfinished jobs?

ga_dashboard/backend/ga_tools.py
```diff
@@ -120,16 +121,18 @@ def extract_data(config_data: dict, has_slurmAdmin: bool, cluster_info) -> pd.Da
...
-    ###### If they are we can maybe create a different WM object with them and run convert2dataframe() command
-
+    unfin_jobs_service = UnfinishedJobsService(config_data)
 
     ### Turn usage logs into DataFrame
-    WM.convert2dataframe()
+    WM.raw_logs_to_df()
+
+    # get unfinished jobs from db and pick those that have now finished using sacct command by job id
+    finished_jobs_df = unfin_jobs_service.sync_unfinished_jobs()
```
Llannelongue: `unfinished` may be clearer than `unfin` in the long run?

ga_dashboard/backend/workload_manager/slurm.py
```diff
@@ -466,3 +448,51 @@ def clean_logs_df(self):
...
+        """
+        Concatenate the existing logs dataframe with a new one, for example when we want to add finished jobs to previously-fetched logs.
+        :param new_logs_df: [pd.DataFrame] new logs dataframe to concatenate with the existing one.
+        """
+        if self.logs_df is None:
+            raise ValueError("logs_df is not initialised. Run pull_logs() and raw_logs_to_df() first.")
+        
+        self.logs_df = pd.concat([self.logs_df, new_logs_df], ignore_index=True)
+
+    def select_distinct_unfinished_jobs(self):
```
Llannelongue, line 462: Can you just explain why we need to do a special aggregation for these jobs?

ga_dashboard/backend/workload_manager/slurm.py
```diff
@@ -466,3 +448,51 @@ def clean_logs_df(self):
...
+
+        def pick_unfinished_state(states: list):
+            for state in states:
+                if state in UNFINISHED_STATES:
+                    return state
+            return states.iloc[0]
+
+        df['JobID_base'] = df['JobID'].astype(str).str.split('.').str[0]
+
+        grouped_by_job = df.groupby('JobID_base').agg(
```
Navirah, line 483: @Llannelongue So the reason why I do this group by here is because in `ga_tools.py` I intend to filter out and save unfinished jobs before it calls `clean_logs_df()` (which in a normal flow performs this kind of grouping). This is to avoid unnecessary calculations that we don't need for unfinished jobs.

ga_dashboard/backend/services/unfinished_jobs_service.py
Navirah: I think the best way to resolve this is to initialize the DBSettings once in main_backend() in ga_tools() and then pass instance in different functions/classes. 

I am making these changes as part of the CI API Issue, hope that is fine.

ga_dashboard/backend/ga_tools.py
Navirah: Agreed, will make this change.

ga_dashboard/backend/services/unfinished_jobs_service.py
Navirah: Yes, I see how that can be confusing, might be best to remove this actually.

ga_dashboard/backend/services/unfinished_jobs_service.py
Navirah: We weren't pulling the 'End' field until now, included this to make it work with the dummy sacct data. But going forward we'll always have 'End'. I would want to keep this for now until we process the CSD3 data we have. Will remove the 'state' bit eventually

ga_dashboard/backend/services/unfinished_jobs_service.py
Navirah: Makes sense. Will make this change.

ga_dashboard/backend/services/unfinished_jobs_service.py
Llannelongue: Right, makes sense. Can you add a comment that this is a temporary workaround for retrocompatibility? So we know to remove it at some point.

ga_dashboard/backend/services/unfinished_jobs_service.py
Llannelongue: Yeah or clarify in the comments

ga_dashboard/backend/workload_manager/slurm.py
Llannelongue: Right got it thanks. Am I right to think that once these finish, you will pull their logs again with acct, and they will get aggregated through the `clean_logs` function?

Navirah: Correct, in the next iteration they'll be picked up from the db and checked for completion using sacct command (`sync_unfinished_jobs()` in `unfinished_jobs_service.py`). Whatever gets picked from here then gets concatenated with the others before `clean_logs_df()` is called.

ga_dashboard/backend/services/unfinished_jobs_service.py
Navirah: Added a comment!

Issue #100, opened 2026-01-24
Use real time carbon intensity for the UK


Llannelongue:
At the bottom, we just enter a carbon intensity value in the config file and use it for all jobs. 

Instead, we should use real time carbon intensity values, in the UK at least, using https://carbonintensity.org.uk. Their API is really easy to use and quite reliable. For an example of this API being used, see https://github.com/GreenScheduler/cats

Default CI value would stay as a fallback option or for users in countries that don't have such a tool. 

Also good to keep it modular to allow for users to submit their own piece of code calling their local API. 

In the future, we could try to engage with electricitymap's own API on this. 

Quite important to be implemented I think.

Navirah:
From discussions with Loïc:

- Carbon Intensity API gives us CI for every 30min interval. These can be interpolated to find specific CI values for jobs that start/end in between these intervals (As done in CATS). We need to check if these CI values are the average over the 30min period or CI at the beginning of that period. 
- The scope of the current Issue would deal with CI values averaged for the entire day rather than taking a granularity approach. 
- Once #133 is done, The CI values corresponding to RUNNING jobs (days when the job runs) must be stored as cache or in a table in the DB to be fetched upon completion of said RUNNING jobs. If these stored CI values are no longer of any use, they must be deleted.
- A post code must be added to _cluster_info.yaml_ to be used during the API call.
- If at any point the call fails or information is unavailable, we must fall back to the hard coded values. (It is important to highlight this during the run) 

Others
- For historical logs, the CI values can be fetched at 1 week intervals (since the API has a 14 day limit). This can be implemented along with Historical logs time-split Issue.

PR #147, opened 2026-03-16
Carbon Intensity API Implementation

Navirah:
Issue #100 

1. API Service added: generic service to call an API using base url, endpoint, api key, headers
2. Carbon Intensity Service implemented: 
           - JobEmissionRecord class to store per job, per day data - also to calculate carbon emission
           - calling Carbon Intensity API in 13 day chunks
           - Calculating, fetching, and storing average CI 
           - Storing n-1 days CI in DB (for today's date average may change)
3. Postcode added to cluster_info.yaml (picking the first 3 chars from it)

Here's a quick overview of the flow:

Screenshot 2026-03-16 at 14 34 04

### DIFF
--- a/configuration/examples/cluster_info__demo.yaml
+++ b/configuration/examples/cluster_info__demo.yaml
@@ -51,6 +51,7 @@ partitions:
     TDP_CPU: 8.8 # from https://ark.intel.com/content/www/us/en/ark/products/91767/intel-xeon-processor-e52650-v4-30m-cache-2-20-ghz.html
 PUE: 1.15 # > 1
 CI: 231.12 # carbon intensity, in gCO2e/kWh
+postcode: "CB1"
 energy_cost:
   cost: 0.34 # in currency/kWh
   currency: "£"
--- a/configuration/examples/cluster_info__demo.yaml
+++ b/configuration/examples/cluster_info__demo.yaml
@@ -51,10 +51,15 @@ partitions:
     TDP_CPU: 8.8 # from https://ark.intel.com/content/www/us/en/ark/products/91767/intel-xeon-processor-e52650-v4-30m-cache-2-20-ghz.html
 PUE: 1.15 # > 1
 CI: 231.12 # carbon intensity, in gCO2e/kWh
-postcode: "CB1"
 energy_cost:
   cost: 0.34 # in currency/kWh
   currency: "£"
+
+# postcode below has to be compatible with the Carbon Intensity API used. 
+# For now, only compatible with UK postcodes (first 3 letters). 
+# If data centre is based outside of UK, leave postcode empty (CI key above will be used instead).
+postcode: "CB1" # [str] Postcode of the data centre
+
 #
 # Below are optional parameters if the html output is used.
 # HTML tags can be used
--- a/configuration/templates/cluster_info.yaml
+++ b/configuration/templates/cluster_info.yaml
@@ -21,6 +21,7 @@ partitions: # a list of the different partitions on the cluster
   # You can keep adding partitions to this
 PUE: <1.67> # [number > 1] Power Usage Effectiveness of the facility
 CI: <467> # [number] average carbon intensity of the geographic location, in gCO2e/kWh
+postcode: <Postcode> # [str] postcode of the HPC system - if HPC is based outside of UK, please leave this empty
 energy_cost:
   cost: <0.34> # [number] in currency/kWh
   currency: "<£>" # [str]
--- a/configuration/templates/cluster_info.yaml
+++ b/configuration/templates/cluster_info.yaml
@@ -21,7 +21,7 @@ partitions: # a list of the different partitions on the cluster
   # You can keep adding partitions to this
 PUE: <1.67> # [number > 1] Power Usage Effectiveness of the facility
 CI: <467> # [number] average carbon intensity of the geographic location, in gCO2e/kWh
-postcode: <Postcode> # [str] postcode of the HPC system - if HPC is based outside of UK, please leave this empty
+postcode: <Postcode> # [str] Postcode of the data centre, has to be compatible with the Carbon Intensity API used. For now, only compatible with UK postcodes (first 3 letters). If data centre is based outside of UK, leave this empty (CI key above will be used instead).
 energy_cost:
   cost: <0.34> # [number] in currency/kWh
   currency: "<£>" # [str]
--- a/configuration/templates/cluster_info.yaml
+++ b/configuration/templates/cluster_info.yaml
@@ -21,10 +21,14 @@ partitions: # a list of the different partitions on the cluster
   # You can keep adding partitions to this
 PUE: <1.67> # [number > 1] Power Usage Effectiveness of the facility
 CI: <467> # [number] average carbon intensity of the geographic location, in gCO2e/kWh
-postcode: <Postcode> # [str] Postcode of the data centre, has to be compatible with the Carbon Intensity API used. For now, only compatible with UK postcodes (first 3 letters). If data centre is based outside of UK, leave this empty (CI key above will be used instead).
 energy_cost:
   cost: <0.34> # [number] in currency/kWh
   currency: "<£>" # [str]
+
+# postcode below has to be compatible with the Carbon Intensity API used. 
+# For now, only compatible with UK postcodes (first 3 letters). 
+# If data centre is based outside of UK, leave postcode empty (CI key above will be used instead).
+postcode: <Postcode> # [str] Postcode of the data centre
 #
 # Below are optional parameters if the html output is used.
 # HTML tags can be used
--- a/ga_dashboard/backend/services/carbon_intensity_service.py
+++ b/ga_dashboard/backend/services/carbon_intensity_service.py
@@ -64,10 +64,10 @@ class CarbonIntensityService:
             raise ValueError("Postcode not found. Cannot fetch CI data.")
 
         try: 
-            # Get CI data in 7-day chunks to avoid API limits
+            # Get CI data in 13-day chunks to avoid API limits (CI API limit is 14)
             chunk_start = from_date
             while chunk_start <= to_date:
-                chunk_end = min(chunk_start + timedelta(days=7), to_date)
+                chunk_end = min(chunk_start + timedelta(days=13), to_date)
 
                 # Formatting as 2026-02-20T00:00Z
                 start_str = chunk_start.replace(hour=0, minute=0).strftime('%Y-%m-%dT%H:%MZ')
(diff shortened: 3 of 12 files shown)


ga_dashboard/backend/services/carbon_intensity_service.py
```diff
@@ -0,0 +1,182 @@
...
+class CarbonIntensityService:
+    def __init__(self, postcode: str, db_params: DBSettings, base_url: str ="https://api.carbonintensity.org.uk/regional/intensity/", api_key: str = None):
+        self.api_service = APIService(base_url, api_key)
+        self.source = base_url.split('/')[2]
+        self.postcode = postcode
+        self.db_params = db_params
+
+    def fetch_CI_data(self, from_date: datetime, to_date: datetime) -> pd.DataFrame:
+        """
+        Fetch CI data in chunks of 7 days. Each chunk is fetched with a single API call.
```
ens-lgil: Is there discrepancies between the 7 days here and the 13 days few line below ?

ga_dashboard/backend/services/carbon_intensity_service.py
```diff
@@ -0,0 +1,182 @@
...
+        except Exception as e:
+            print(f"Error occurred while calling CarbonIntenistyAPI: {e}")
+            return pd.DataFrame()
+        
+        if not all_data:
+            return pd.DataFrame()
+        
+        ci_df = pd.DataFrame(all_data)
+        ci_df['intensity_value'] = ci_df['intensity'].apply(
+            lambda x: x.get('forecast') # Regional carbon intensity API provides CI as forecast
```
ens-lgil, line 121: Could we provide an example in comments about the format of the `forecast` column ?

ga_dashboard/database/ga_db.sql
```diff
@@ -105,6 +105,49 @@ CREATE TABLE public.ga_user (
...
+ALTER TABLE ONLY public.ga_unfinished_jobs
+    ADD CONSTRAINT ga_unfinished_jobs_pkey PRIMARY KEY (job_id);
+
+--
+-- Name: carbon_intensity_data; Type: TABLE; Schema: public; Owner: -
+--
+
+DROP TABLE IF EXISTS public.carbon_intensity_data; -- Needed due to postgres caching
+
+CREATE TABLE public.carbon_intensity_data (
```
ens-lgil, line 136: Do we need to store the postcode as well ?
If so, we will need to add the postcode in the fetch methods in the `services/carbon_intensity_service` module.

ga_dashboard/backend/services/carbon_intensity_service.py
```diff
@@ -0,0 +1,182 @@
...
+            ]
+        
+        ci_data_from_db = pd.DataFrame()
+        ci_data_from_api = pd.DataFrame()
+
+        try:
+            # Trying to pick previously saved data from DB
+            ci_data_from_db = self.fetch_stored_average_CI(dates_list)
+
+            # Find dates for which data is not available in the DB
```
ens-lgil, line 187: Typo ?
> Find dates for which data is ~not~ available in the DB

ga_dashboard/backend/services/carbon_intensity_service.py
Navirah: Yes I was doing 7 days initially, but changed it to 13, will fix this. Thanks for pointing it out!

ga_dashboard/backend/services/carbon_intensity_service.py
Navirah: Sure!

ga_dashboard/database/ga_db.sql
Navirah: I had initially included post code as a column, but after our discussion in the meeting last week, I gathered we don't need to store post code since the HPC system is most probably housed in a single post code. 

Might need to check how to save this post code to show on the dashboard though🤔
Can we show variables on Grafana that we're not saving in the DB? Can we at run time change the Grafana json?

ga_dashboard/backend/services/carbon_intensity_service.py
Navirah: Here I am trying to filter out the dates for which the avg is not stored in the db, then calling the api's for those dates.

ga_dashboard/backend/services/carbon_intensity_service.py
Navirah: I've added the link to the doc: https://carbon-intensity.github.io/api-definitions/#get-regional-intensity-from-to

ga_dashboard/database/ga_db.sql
ens-lgil: Could it be having an additional table storing some information about the server ?

configuration/examples/cluster_info__demo.yaml
```diff
@@ -51,6 +51,7 @@ partitions:
     TDP_CPU: 8.8 # from https://ark.intel.com/content/www/us/en/ark/products/91767/intel-xeon-processor-e52650-v4-30m-cache-2-20-ghz.html
 PUE: 1.15 # > 1
 CI: 231.12 # carbon intensity, in gCO2e/kWh
+postcode: "CB1"
```
Llannelongue: (use same comment as in the template file)

configuration/templates/cluster_info.yaml
```diff
@@ -21,6 +21,7 @@ partitions: # a list of the different partitions on the cluster
   # You can keep adding partitions to this
 PUE: <1.67> # [number > 1] Power Usage Effectiveness of the facility
 CI: <467> # [number] average carbon intensity of the geographic location, in gCO2e/kWh
+postcode: <Postcode> # [str] postcode of the HPC system - if HPC is based outside of UK, please leave this empty
```
Llannelongue:
```suggestion
postcode: <Postcode> # [str] Postcode of the data centre, has to be compatible with the Carbon Intensity API used. For now, only compatible with UK postcodes (first 3 letters). If data centre is based outside of UK, leave this empty (CI key above will be used instead).
```

ga_dashboard/backend/services/carbon_intensity_service.py
```diff
@@ -0,0 +1,183 @@
...
+# ------------------------------------------------------------------
+
+import pandas as pd
+from datetime import datetime, timedelta
+
+from ga_dashboard.backend.services.api_service import APIService
+from ga_dashboard.backend.services.database_service import DBSettings, DatabaseService
+from ga_dashboard.database.table_col_definitions import CARBON_INTENSITY_DATA_COLUMNS
+
+class JobEmissionRecord:
```
Llannelongue, line 13: Can you add a brief description of the classes in this file?

ga_dashboard/backend/services/carbon_intensity_service.py
```diff
@@ -0,0 +1,183 @@
...
+
+        :param records: time period slices of the job, each with duration and CI
+        :param energy_per_hr: energy consumed per hour (total_energy / total_duration)
+        :return: total carbon emission in gCO2
+        """
+        if not records:
+            print("calc_carbon_emission(): No records provided.")
+            return 0.0
+
+        weighted_CI = sum(r.hours_of_work * r.carbon_intensity for r in records if r.carbon_intensity is not None)
```
Llannelongue, line 52: Not for this PR: but can we add an issue with the `test` label to make sure that we have some unit test checking that all these functions return the numerical values expected?

ga_dashboard/backend/services/carbon_intensity_service.py
```diff
@@ -0,0 +1,183 @@
...
+
+        return energy_per_hr * weighted_CI
+
+class CarbonIntensityService:
```
Llannelongue, line 56: It would be good to make it clear that this is what is specific to the UK API (or if it's not this whole class, what is API specific, and what can easily be adapted to other APIs).

ga_dashboard/backend/services/carbon_intensity_service.py
```diff
@@ -0,0 +1,183 @@
...
+        except Exception as e:
+            print(f"Error occurred while calling CarbonIntenistyAPI: {e}")
+            return pd.DataFrame()
+        
+        if not all_data:
+            return pd.DataFrame()
+        
+        ci_df = pd.DataFrame(all_data)
+        ci_df['intensity_value'] = ci_df['intensity'].apply(
+            lambda x: x.get('forecast') # Regional carbon intensity API provides CI as forecast
```
Llannelongue, line 121: A question about the API itself: Are the values returned the forecasts made by their model, or is this replaced by actual CI values for past dates?

Llannelongue (commented):
Looks good to me pending these minor comments.

ga_dashboard/database/ga_db.sql
Navirah: (From discussion on Dashboard meeting 19 March)

For now we can add a line on the UI saying CI graph reflects CI per day average for the location of the HPC. However, this would not work for non-UK based HPC (but then, non-UK based HPC systems would not have a CI graph, unless required API changes are made)

Can create an Issue to create table for storing cluster info for later versions of dashboard.

ga_dashboard/backend/services/carbon_intensity_service.py
Navirah: Alright, I'll make a note of this.

ga_dashboard/backend/services/carbon_intensity_service.py
Navirah: Added.

configuration/examples/cluster_info__demo.yaml
Navirah: Done

ga_dashboard/backend/services/carbon_intensity_service.py
Navirah: Added.

ga_dashboard/backend/services/carbon_intensity_service.py
Navirah: What I understand from reading the documentation is that:

- Regional CI is reported as _forecast_ because it is essentially a result from the machine learning models and not reported using an exact meter (unlike overall intensity, which is measured by NESO). Therefore, the regional API does not report the _actual_ intensity unlike `https://api.carbonintensity.org.uk/intensity`

- Every 30 mins the _forecast_ updates for near future, but the past ones remain consistent. 

Its very interesting to see how they're doing the predictions in their methodology (attaching it here) [neso-ci-regional-methodology_v2.pdf](https://github.com/user-attachments/files/26218508/neso-ci-regional-methodology_v2.pdf)

Llannelongue: Interesting, thanks!

PR #153, opened 2026-03-27
Historical logs pull in batches 

Navirah:
Issue #139 

- Encapsulated the following in a class called `LogsDataProcessor`:
         - `main_backend()`
         - `extract_data()`
         - `enrich_data()`
         - `summarise_data()`
         - `process_and_store()`
(now both historical run and daily run process logs using this class)
         
 - Created 2 functions - `run() `and `batch_run()` - for running the data processing pipeline (`extract_data() -> enrich_data() -> summaries_data() -> process_and_store()`) in one go and in batches respectively.

- Added deletion and garbage collection for large data frames in both data processing pipeline and batch run.

 - Updated printed logs that are displayed when running `run_green_algorithms_on_historical_logs` and `batch_run`
 
Screenshot 2026-03-27 at 15 50 21

_Minor changes:_

 - `concat_dataframes()` util added to concatenate 2 data frames after checking for completely empty data frames - done this to resolve **Future Warnings** that were reported when running `main_backend()` ( occurring because of the line `daily_avg = pd.concat([ci_data_from_db, ci_data_from_api]).reset_index(drop=True)` in `carbon_intensity_service.py`

- resolved some more **Future Warnings** caused by `groupby()` lines in `summaries_data()`

- Commented some print messages that reported when mock data is used - these can be uncommented if required during development

PR #159, opened 2026-03-31
Removing Unfinished Jobs + logs processing clean up

Navirah:
**Contains commits from the PR #153 - That needs to be merged before this one.** (done this way to avoid conflict)

### Changes
- Removed UnfinishedJobsService - no longer needed. The SLURM sacct command picks jobs that are active within `[starttime, endtime]` period - no matter their status.
- Filtering finished jobs in slurm.py - Unfinished jobs get picked by sacct, its better to remove them before cleaning to avoid unnecessary df.apply processing
- Added a test csv file with multiple cases of slurm logs - tried to make it as realistic as possible.

### Testing it
ASSUMING: the script runs at **4am on 31st March**
SETUP: a csv file with dummy jobs - `tests/testdata/slurm_logs_many_cases.csv`
- Jobs fully contained within the period – started and finished during the 24 hours.
- Jobs that started before the period but finished within it – partially overlapping at the start.
- Jobs still running at the end of the period – either started before or during the window.
- Various states – COMPLETED, FAILED, CANCELLED, TIMEOUT, RUNNING.

| Category                                    | Count | JobIDs                                         |
| ------------------------------------------- | ----- | ---------------------------------------------- |
| Fully contained in the window               | 6     | 700001, 700002, 700003, 700006, 700011, 700012 |
| Started before, ended within the window     | 2     | 700004, 700005                                 |
| Started before, still running at window end | 2     | 700009, 700010                                 |
| Started during window, still running        | 2     | 700007, 700008                                 |
| **Total jobs overlapping window**           | 12    | All jobs                                       |

**Verified that 8 out of 12 jobs were processed - correct as per logic.**

Some things to note:
1. If a job runs for multiple days, it gets picked up everyday but gets removed each day unless it has finished.
2. Carbon intensity is only fetched for jobs that have finished. So lets say a job has been running for the past 4 days, the CI for past 4 days will only be fetched and avg. will be calculated and stored once this job completed - done this way for simplicity

P.S. it was actually quite easy to remove Unfinished jobs - good to see the Service-based structure do its job - can definitely be improved.


ga_dashboard/backend/workload_manager/slurm.py
```diff
@@ -5,7 +5,9 @@
 
 from ga_dashboard.backend.helpers import utils
 from ga_dashboard.backend.services.sacct_service import SacctService
-from ga_dashboard.backend.services.unfinished_jobs_service import UNFINISHED_STATES, UnfinishedJobsService
+
+## States from SLURM documentation: https://slurm.schedmd.com/job_state_codes.html (as of 10 Feb 2026)
+UNFINISHED_STATES = ['PENDING','RUNNING','SUSPENDED','UNKNOWN','PREEMPTED']
```
ens-lgil: It could be included in the `SlurmBase` class (not important)

scripts/run_green_algorithms_on_historical_logs.py
```diff
@@ -18,10 +20,18 @@
                             metavar='CONFIG_FILE', dest='config')
     argparser.add_argument("--db_pass", help='Database password (optional, to avoid entering it in the prompt).', \
                             metavar='DB_PASS', dest='db_pass')
+    argparser.add_argument("--batch_size", help='Batch size in number of days. Default size is 30 days',\
```
ens-lgil: Could we create a `default_batch_size=30` variable to use in the `Default size is {default_batch_size} days` and then below in the `if args.batch_size is None:` block ?
It's just that if we change it in the future, we will need to change it only in 1 place.

Navirah: I thought since we're using batch_size only in a single script, we wouldn't need an extra variable.

ga_dashboard/backend/workload_manager/slurm.py
Navirah: Yeah makes more sense, added it. commit [fba01b8](https://github.com/Cambridge-Sustainable-Computing-Lab/Green-Algorithms-HPCdashboard/pull/159/commits/fba01b880b946d3b0ff3a31288b130feb146ade2)

scripts/run_green_algorithms_on_historical_logs.py
ens-lgil: Yes, the idea would be to use a local variable as you have the value `30` written several times in the script.

Navirah: Makes sense, will add a local variable.

Navirah: I've this as part of historical data in batches PR - #153 

commit [8535efe](https://github.com/Cambridge-Sustainable-Computing-Lab/Green-Algorithms-HPCdashboard/pull/153/commits/8535efe15cbc99e41348eb0fd3ac8d627222dbc1)

Llannelongue (approved):
Looks good to me! Good job on the thorough testing.

PR #176, opened 2026-07-14
Separating ga_core and adding a test suite for ga_tools

Navirah:
This PR does a couple of things that are inter-related:
1. It removes the `ga_core` modules from the dashboard and instead uses the [`Green-Algorithms-core` package](https://github.com/Cambridge-Sustainable-Computing-Lab/Green-Algorithms-core). 
2. Implements a test suite to perform basic tests on the `ga_tools` module to check end-to-end data pipeline. This helps test that the integration of `ga_core` package is done correctly.
3. Removes older tests that are no longer usable due to major changes in the code structure. 

### Using `ga_core` package
- `LogsDataProcessor` in `ga_tools` now uses modules from the `ga_core` package to extract and enrich logs data.
- `pyproject.toml` now contains `green-algorithms-core` as a dependency - pulling the package straight from GitHub (currently pointing to `main`, eventually must be updated with the exact release).
- `requirements.txt` also has `ga_core` as a dependency.

### Changes to older code
- updated `simulate_mock_jobs()` utility function to include start and end dates for each mock job.
- `pickle_it.py` refactored to accept `workload_manager`as command line argument and mock jobs accordingly - this module is useful to quickly extract enriched data and update tests when the core functionality changes.
- csv files containing examples of raw logs moved inside a 'slurm' folder. 
- Outdated mock files and tests deleted.

### New test structure and test suite
- tests folder structure improved to make it less slurm specific.
- test data better organised
- implemented `conftest.py` and `helpers.py` for configuring test cases and defining test helpers respectively.
- `test_ga_tools.py` implemented to test the basic data pipeline to ensure the data is correctly processed after importing the new `ga_core` package.

NOTE: Github workflows to automatically run test cases for every PR and merge into main/dev will be implemented once all refactor changes are merged into main and dev. If we do it right now, the older test cases in main will fail.

For now attaching a screenshot of successful test case run:
Screenshot 2026-07-14 at 14 46 50


tests/pickle_it.py
```diff
@@ -1,10 +1,40 @@
-# Run this from GA4HPCdashboard/ga_dashboard directory:
-# python tests/pickle_it.py
+# ------------------------------------------------------------------
+# Mock Data Generator for GA4HPC Dashboard Tests
```
Llannelongue: Green Algorithms Dashboard no?

tests/pickle_it.py
```diff
@@ -1,10 +1,40 @@
...
+# This script generates a synthetic enriched logs dataset. i.e. with energy consumption and carbon footprint data,
```
Llannelongue: I'm confused why a script called `pickle_it.py` mainly "generates a synthetic enriched logs dataset"? (and where `simulate_mock_jobs()` should sit)

Llannelongue (commented):
Looks good to me overall, I'm confused about what `pickle_it.py` does and why it's named like that? 

It would also be useful to understand test coverage: i.e what is being tested and what still needs tests.

tests/pickle_it.py
Navirah: Yes my bad, have been working on both simultaneously😅 Will change it.

tests/pickle_it.py
Navirah: I'm assuming this file was named `pickle_it.py` because it performs pickling of the mock data (Pickling here refers to converting an in-memory object into a byte stream so it can be saved to a file)

It kind of made sense to me so I left it as such, but you're right that it is unclear. Will rename it (do you have any suggestions?)

Although currently its being used only to mock jobs, in the future we can implement other functions that save other objects for testing - such as CI average data, mock users list etc.

Navirah:
> It would also be useful to understand test coverage: i.e what is being tested and what still needs tests.

I was reading how to calculate code coverage in GitHub workflows, will add it there once I create a workflow for this repo.

tests/helpers.py
```diff
@@ -0,0 +1,30 @@
+from datetime import date
+import pandas as pd
+
+
+def align_expected_dtypes(expected: pd.DataFrame, result: pd.DataFrame) -> pd.DataFrame:
+    """
+    Loading CSV filesloses dtype info - every column comes back as a string.
```
ens-lgil: Typo `filesloses`

tests/pickle_it.py
ens-lgil: Do we need this script on top of the mockup data files we already have ?

ens-lgil (approved):

tests/helpers.py
Navirah: Thanks will fix this! 
For some reason I'm not able to comment on your other comment, so responding here:

> Do we need this script on top of the mockup data files we already have ?

This script is for cases when the data changes, for example when new fields get added. Using this we can generate new mock files to use for test cases.

tests/pickle_it.py
Llannelongue: I would be cautious of doing something jut for the hypothetical case we may need it later. If it's only to generate mock data, wouldn't it make more sense to add that to the helpers or something like that?

Navirah: Right, that makes sense. I'll move it.

Navirah: I've removed `pickle_it.py` but kept the utility function to create mock file. Commit [2175c1e](https://github.com/Cambridge-Sustainable-Computing-Lab/Green-Algorithms-HPCdashboard/pull/176/commits/2175c1e4f74d48de2f69871ed00492d92d1c1dc3)

Llannelongue (approved):

